PR #4, opened 2020-05-21
Queue list UI

bugsalexander:
"responsive as heck" –stanley liu 2020

### DIFF
--- a/packages/app/components/Queue/QueueCard.tsx
+++ b/packages/app/components/Queue/QueueCard.tsx
@@ -1,9 +1,11 @@
-import { Card, Row, Avatar, Tag } from "antd";
+import { Card, Row, Col, Avatar, Tag } from "antd";
 import { UserOutlined } from "@ant-design/icons";
 import styled from "styled-components";
 
 const HorizontalCard = styled(Card)`
   margin-bottom: 8px;
+  padding-left: 8px;
+  padding-right: 8px;
 `;
 
 const Text = styled.div`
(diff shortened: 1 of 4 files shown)


packages/app/components/Queue/QueueCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+`;
+
+const CenterRow = styled(Row)`
+  align-items: center;
+`;
+
+interface QueueCardProps {
+  rank: number;
+  name: string;
+  questionType: string;
```
NEUDitao, line 35: should this be an enum/type of some type? i know we don't have concrete question types yet, but I think it's better to future proof it now

packages/app/components/Queue/QueueList.tsx
```diff
@@ -0,0 +1,110 @@
...
+          <Col span={2}>
+            <HeaderText>wait</HeaderText>
+          </Col>
+          <Col xs={0} lg={2}>
+            <StatusText>status</StatusText>
+          </Col>
+        </CenterRow>
+      </HeaderCard>
+
+      <QueueCard
```
NEUDitao, line 73: we could get rid of the dummy data if we just finished the project rq smh my head

dajinchu:
Vercel can deploy frontend now! https://office-hours-git-queue-list.sandboxneu.now.sh/queue

willstenzel:
Screen Shot 2020-05-24 at 10 09 37 AM
Looks solid on mobile. Nice work

packages/app/components/Queue/QueueCard.tsx
liustanley: yeah def, we made an enum in #5 but i just moved it into /common for this PR

PR #5, opened 2020-05-24
Join queue

bugsalexander:
**this should only be merged to master after #4 is merged**

add the form page UI + state management for joining the queue.

note: we couldn't find an easy way to check for (and warn) the user when they try to navigate away from the join page (to warn them they would lose their spot in the queue). our solution to this was to throw an error in the callback, which is not optimal (lmao). 

any suggestions welcome.

### DIFF
--- a/packages/app/components/Queue/QuestionForm.tsx
+++ b/packages/app/components/Queue/QuestionForm.tsx
@@ -1,7 +1,9 @@
 import { Button, Input, Radio, Alert } from "antd";
 import styled from "styled-components";
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { RadioChangeEvent } from "antd/lib/radio";
+import Router from "next/router";
+import Link from "next/link";
 
 const Container = styled.div`
   max-width: 960px;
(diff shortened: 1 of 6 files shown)


dajinchu:
Other than being jank af, is there any functionality that doesn't work when throwing in the callback? Maybe browser back/forward buttons?

An alt solution: https://github.com/zeit/next.js/pull/4474 see: "cancel the navigation if e.defaultPrevented === true"

liustanley:
I was actually trying to prevent the use of browser back/forward buttons, since I figured we wouldn't want a student to accidentally go back or close the window and lose their reserved spot in queue. Is this a behavior we would want? Or should we only be adding the warning to Links?

willstenzel:
If we wanted to give the user a warning whenever they might leave the page would it work to use a beforeUnload hook to trigger a warning?
https://www.npmjs.com/package/react-beforeunload

However, you bring up a good question which I'm not sure the best solution to. I'm also curious in the senario where someone joins the queue using their phone and then wants to switch over to using their computer, would they loose their spot since they would have to close the browser window on their phone?

liustanley:
> If we wanted to give the user a warning whenever they might leave the page would it work to use a beforeUnload hook to trigger a warning?
> https://www.npmjs.com/package/react-beforeunload
> 
> However, you bring up a good question which I'm not sure the best solution to. I'm also curious in the senario where someone joins the queue using their phone and then wants to switch over to using their computer, would they loose their spot since they would have to close the browser window on their phone?

I looked at the beforeUnload hook and it works well, but it only triggers when the user is attempting to close the webpage. I think that multiple devices is a good point, and it might make more sense to not automatically close student draft questions when they leave the page.

The behavior that makes sense to me is that the only way for a student to close their draft themselves is through the "Leave Queue" button in the drafting page. If they leave the drafting page without pressing "Leave Queue", their unfinished question will remain at the top of the page with how much time they have remaining to fill out the question. That way, no matter what happens they can finish their draft within those five minutes or so, and if they don't finish within the time frame the backend will close their question.

Any thoughts on this?

bugsalexander:
Alternative solution for this issue: when a student navigates away from this page, the in-progress ticket minimizes to be a card on top of the page in the bottom right corner, with a countdown over it that can then be clicked on to return to this page, with their progress saved. if logging in on another device, the in-progress ticket appears as the same card in the bottom left.

willstenzel:
I like the idea of having the ticket be minimized when they navigate away from the page I think that's the best UX if we can pull it off. 

We'll probably want to consult some students and the team on if we need a countdown and if so how long. Just want to make sure that the countdown doesn't cause any additional stress to be put on the student.

liustanley:
I agree, the minimized card sounds good. I just removed the janky navigation prevention, so this PR is ready for review if someone wants to take a look.

packages/app/components/Queue/QuestionForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+        autoSize={{ minRows: 3 }}
+        onChange={onQuestionTextChange}
+      />
+      <QuestionCaption>
+        Be as descriptive and specific as possible in your answer. If your
+        question matches another student’s, your wait time may be reduced.
+      </QuestionCaption>
+
+      <div>
+        <FormButton type={computeQuestionType()} onClick={onClickSubmit}>
```
willstenzel, line 132: Everything looks great. The only change I see that could be made is on this line. 
```suggestion
        <FormButton type="primary" disabled={!questionType || !questionText || questionText === ""} onClick={onClickSubmit}>
```
It might be slightly more readable from a code standpoint and will show the button as disabled in addition to automatically changing the color. You could then remove the `computeQuestionType()` function which took me a second to figure out what it was doing at first.

liustanley: Good catch, I didn't know there was a disabled property on button.

PR #7, opened 2020-05-30
Ical frontend

NEUDitao:
This adds:

- big calendar for the calenlolz
- our first real endpoint
- our first real fixtures
- and showcasing `useSWR` which we should use in the future to fetch data

### DIFF
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -24,7 +24,6 @@ class APIClient {
         e.startTime = new Date(e.startTime);
         e.endTime = new Date(e.endTime);
       });
-      throw Error();
       return data;
     },
   };
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -3,6 +3,7 @@ import {
   GetClubResponse,
   CreateClubParams,
   CreateClubResponse,
+  GetCourseResponse,
 } from "@template/common";
 
 class APIClient {
@@ -15,6 +16,18 @@ class APIClient {
       return (await this.axios.post("/api/club", p)).data;
     },
   };
+
+  course = {
+    get: async (courseId: number): Promise<GetCourseResponse> => {
+      const data = (await this.axios.get(`/api/v1/courses/${courseId}`)).data;
+      data.officeHours.forEach((e: any) => {
+        e.startTime = new Date(e.startTime);
+        e.endTime = new Date(e.endTime);
+      });
+      return data;
+    },
+  };
+
   constructor(baseURL: string = "") {
     this.axios = Axios.create({ baseURL: baseURL });
   }
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -24,6 +24,7 @@ class APIClient {
         e.startTime = new Date(e.startTime);
         e.endTime = new Date(e.endTime);
       });
+      throw Error();
       return data;
     },
   };
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -3,8 +3,8 @@ import {
   GetClubResponse,
   CreateClubParams,
   CreateClubResponse,
-  GetCourseResponse,
 } from "@template/common";
+import { GetCourseResponse } from "../common";
 
 class APIClient {
   private axios: AxiosInstance;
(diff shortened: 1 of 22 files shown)


packages/server/src/api/courseRoutes.ts
```diff
@@ -0,0 +1,39 @@
...
+      return {
+        name: course.name,
+        officeHours: (await course.officeHours).map((e) =>
+          pick(e, ["id", "title", "room", "startTime", "endTime"])
+        ),
+      };
+    },
+    options: {
+      response: {
+        schema: Joi.object({
```
willstenzel, line 24: Would the type `Promise<GetCourseResponse>` cause an error to be thrown even if the schema wasn't defined?

packages/app/pages/schedule.tsx
```diff
@@ -0,0 +1,33 @@
+import { Calendar, momentLocalizer } from "react-big-calendar";
+import moment from "moment";
+import useSWR from "swr";
+import { API } from "@template/api-client";
+import { Result } from "antd";
+
+export default function Schedule() {
+  const { data, error } = useSWR(`api/v1/courses/1`, async () =>
```
willstenzel, line 8: Might be slightly easier to just use the `/v1/courses/{course_id}/` for the routes so we don't have to add the extra `/api/` onto all of them but that's not a big deal.

packages/common/index.ts
```diff
@@ -12,6 +12,17 @@ export type GetClubResponse = Club[];
...
 
+export type GetCourseResponse = {
+  name: string;
+  officeHours: Array<{
+    id: number;
+    title: string;
+    room: string;
+    startTime: Date;
+    endTime: Date;
+  }>;
```
liustanley, line 23: Would it be better to make this object type into an interface? That way the frontend can know what data type to expect inside the array.

packages/server/src/api/courseRoutes.ts
```diff
@@ -0,0 +1,39 @@
...
+export const courseRoutes: ServerRoute[] = [
+  {
+    method: "GET",
+    path: "/api/v1/courses/{course_id}",
+    handler: async (request, h): Promise<GetCourseResponse> => {
+      const course = await CourseModel.findOne(request.params.course_id, {
+        relations: ["officeHours"],
+      });
+      return {
+        name: course.name,
```
willstenzel, line 16: Would we also want to include the `course.id` here?

packages/app/pages/schedule.tsx
NEUDitao: we have to prepend everything with `api` due to some gobblygook

packages/api-client/index.ts
```diff
@@ -15,6 +16,19 @@ class APIClient {
...
   };
+
+  course = {
+    get: async (courseId: number): Promise<GetCourseResponse> => {
+      const data = (await this.axios.get(`/api/v1/courses/${courseId}`)).data;
+      data.officeHours.forEach((e: any) => {
+        e.startTime = new Date(e.startTime);
+        e.endTime = new Date(e.endTime);
+      });
+      throw Error();
```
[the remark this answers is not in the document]
NEUDitao: nope :party_try_not_to_cry:

packages/server/src/api/courseRoutes.ts
NEUDitao: we figured not, since you'll already know the course id if you're querying the backend for it.

packages/app/pages/schedule.tsx
```diff
@@ -0,0 +1,33 @@
...
+    API.course.get(1)
+  );
+
+  if (error)
+    return (
+      <Result
+        status="500"
+        title="Something went wrong, please ask chinese man"
```
bugsalexander, line 16: :thonk:

packages/app/pages/schedule.tsx
```diff
@@ -0,0 +1,33 @@
...
+      />
+    );
+
+  const myEvents =
+    data?.officeHours.map((e) => ({
+      title: e.title,
+      start: e.startTime,
+      end: e.endTime,
+    })) || [];
```
bugsalexander, line 25:
```suggestion
    })) ?? [];
```

PR #8, opened 2020-05-31
Mock api 

bugsalexander:
adds the mock api (minus courses, because eddy/dj worked on that). 

things we should probably do before merge:
- [x] add JOI to routes that have payloads
- [x] finish adding tests for routes
- [x] clean up routes (easier to read)


bugsalexander:
@willstenzel last we talked, we decided that we should eventually go back to using enums, so I went ahead and changed QuestionType from union > enum. It was preventing typechecking.

packages/server/src/api/officeHoursRoutes.ts
```diff
@@ -1,27 +1,99 @@
...
       } else {
-        // todo: we should fix this later
-        return h.response("The course did not exist").code(404) as any;
+        return h.response("The course did not exist").code(404);
+      }
+    },
+  },
+  {
+    method: "GET",
+    path: "/v1/courses/{course_id}/ta/change_status",
```
bugsalexander, line 50: According to the [api docs](https://www.notion.so/API-Docs-aa0846685baf4f8ca5f187cd2c6ef38d), this should be a PATCH, instead of a GET.

willstenzel: Whoops! Thanks for the comment. Just got that changed

packages/api-client/index.ts
```diff
@@ -11,13 +20,94 @@ class APIClient {
...
+    update: async (courseId: number): Promise<TAUpdateStatusResponse> => {
+      const queue = (
+        await this.axios.patch(`/v1/courses/${courseId}/ta/change_status`)
+      ).data;
+      parseQueueDates(queue);
+      queue.questions.forEach((question: any) => parseQuestionDates(question));
+      return queue;
+    },
+  };
+  questions = {
```
NEUDitao, line 51: we're gonna have to prepend each of these endpoints with `/api` in the future, since frontend routing is all being done automatically by Next, prepending with `/api` makes our routing not bugger out

packages/api-client/index.ts
```diff
@@ -11,13 +20,94 @@ class APIClient {
...
+      parseQuestionDates(question);
+      return question;
     },
   };
   constructor(baseURL: string = "") {
     this.axios = Axios.create({ baseURL: baseURL });
   }
 }
 
+function parseOfficeHourDates(officeHour: any): void {
```
NEUDitao, line 97: should these be explicitly `any`? I'm fine with either defining a type for it, or keeping it implicit

packages/server/src/mocks/updateQuestion.ts
```diff
@@ -0,0 +1,39 @@
...
+  creator: {
+    id: 202,
+    name: "Ryan Drew",
+    photoURL:
+      "https://prod-web.neu.edu/wasapp/EnterprisePhotoService/PhotoServlet?vid=CCS&er=93e5370e7fdbdb669f4a75c4de9cfa9ccffbe3e22119d3a51aed1d485d85129e1455cc78b34614367ec6a8a363ca61ed6bd1074bd00c3e6e",
+  },
+  text: "Having trouble with recursion on problem 2",
+  taHelped: {
+    id: 132,
+    name: "Eddie Li",
```
NEUDitao, line 31: i spell it "Eddy" :P

packages/server/src/joi.ts
```diff
@@ -0,0 +1,89 @@
+import Joi from "@hapi/joi";
+import { courseRoutes } from "./api/courseRoutes";
+import { QuestionType } from "@template/common";
+
+// type CoursePartial
+export const CoursePartialSchema = Joi.object({
```
NEUDitao, line 6: for all of the Joi stuff, do we want to let things like `id`, `name` be null? If not, you'll want to add `.options({presence: 'required'})` to each of these ala [here](https://hapi.dev/module/joi/api/?v=17.1.1#anyvalidatevalue-options)

packages/api-client/index.ts
willstenzel: The reason it's an `any` is that its super similar to the the type `OfficeHourBlock` but the dates are strings instead of Date object. We figured since there is Joi validations on the response from the server we can assume that data will be right form. 
Now that I think of it though, I'm not sure what would happen if the server responds with an error?

packages/server/src/joi.ts
willstenzel: I think that the `schema: CourseSchema.options({ presence: "required" }),` in the courseRoutes.ts file bubbles up so it requires non-null for all the objects that it is composed of.

PR #9, opened 2020-06-05
Creates today page and uses MOCK DATA (not fixture) for backend currently

NEUDitao:
We:

- created the today page :) (and kinda on mobile but not really but at least it stacks 🥳 )
- created the endpoint /v1/courses/{course_id}/queue
- moved /v1/courses/{course_id}/ -> same but with schedule at the end
- removed the title header from the calendar because the navbar will have it one day
- made a temp navbar
- made some probably questionable frontend maneuvers. pls guide us senpai
- updated some types but ignore `yarn tsc`

### DIFF
--- a/packages/app/pages/schedule.tsx
+++ b/packages/app/pages/schedule.tsx
@@ -4,7 +4,11 @@ import useSWR from "swr";
 import { API } from "@template/api-client";
 import { Result } from "antd";
 
-export default function Schedule() {
+type ScheduleProps = {
+  viewType: Calendar.Views; // TODO: debug this, hwy isn't it finding the types?
+};
+
+export default function Schedule({ viewType }: ScheduleProps) {
   const { data, error } = useSWR(`api/v1/courses/1`, async () =>
     API.course.get(1)
   );
@@ -27,7 +31,11 @@ export default function Schedule() {
   return (
     <div style={{ height: "400px" }}>
       <h1>{data?.name ?? ""}</h1>
-      <Calendar localizer={momentLocalizer(moment)} events={myEvents} />
+      <Calendar
+        localizer={momentLocalizer(moment)}
+        events={myEvents}
+        defaultView={viewType}
+      />
     </div>
   );
 }
--- a/packages/app/pages/schedule.tsx
+++ b/packages/app/pages/schedule.tsx
@@ -14,10 +14,8 @@ type ScheduleProps = {
 };
 
 export default function Schedule({ viewType }: ScheduleProps) {
-  const { data, error } = useSWR(
-    `api/v1/courses/1/schedule`,
-    async () => API.course.get(1),
-    { revalidateOnFocus: false }
+  const { data, error } = useSWR(`api/v1/courses/1/schedule`, async () =>
+    API.course.get(1)
   );
 
   if (error)
--- a/packages/app/pages/schedule.tsx
+++ b/packages/app/pages/schedule.tsx
@@ -1,16 +1,23 @@
-import { Calendar, momentLocalizer } from "react-big-calendar";
+import { Calendar, momentLocalizer, View } from "react-big-calendar";
 import moment from "moment";
 import useSWR from "swr";
 import { API } from "@template/api-client";
 import { Result } from "antd";
+import styled from "styled-components";
 
 type ScheduleProps = {
-  viewType: Calendar.Views; // TODO: debug this, hwy isn't it finding the types?
+  viewType: View;
 };
 
+const CalendarContainer = styled.div`
+  height: 70%;
+`;
+
 export default function Schedule({ viewType }: ScheduleProps) {
-  const { data, error } = useSWR(`api/v1/courses/1/schedule`, async () =>
-    API.course.get(1)
+  const { data, error } = useSWR(
+    `api/v1/courses/1/schedule`,
+    async () => API.course.get(1),
+    { revalidateOnFocus: false }
   );
 
   if (error)
@@ -29,12 +36,12 @@ export default function Schedule({ viewType }: ScheduleProps) {
   // TODO: frontend guru pls cleanup height <3, month view is crunked :)
   // esp on mobile
   return (
-    <div style={{ height: "70%" }}>
+    <CalendarContainer>
       <Calendar
         localizer={momentLocalizer(moment)}
         events={myEvents}
         defaultView={viewType}
       />
-    </div>
+    </CalendarContainer>
   );
 }
--- a/packages/app/pages/schedule.tsx
+++ b/packages/app/pages/schedule.tsx
@@ -5,13 +5,13 @@ import { API } from "@template/api-client";
 import { Result } from "antd";
 import styled from "styled-components";
 
+const ScheduleCalendar = styled(Calendar)`
+  height: 70vh;
+`;
+
 type ScheduleProps = {
   viewType: View;
 };
-
-const CalendarContainer = styled.div`
-  height: 70%;
-`;
 
 export default function Schedule({ viewType }: ScheduleProps) {
   const { data, error } = useSWR(
@@ -33,15 +33,12 @@ export default function Schedule({ viewType }: ScheduleProps) {
       start: e.startTime,
       end: e.endTime,
     })) ?? [];
-  // TODO: frontend guru pls cleanup height <3, month view is crunked :)
-  // esp on mobile
+
   return (
-    <CalendarContainer>
-      <Calendar
-        localizer={momentLocalizer(moment)}
-        events={myEvents}
-        defaultView={viewType}
-      />
-    </CalendarContainer>
+    <ScheduleCalendar
+      localizer={momentLocalizer(moment)}
+      events={myEvents}
+      defaultView={viewType}
+    />
   );
 }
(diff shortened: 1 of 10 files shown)


packages/app/pages/today.tsx
```diff
@@ -0,0 +1,91 @@
...
+        title={staffList.map((staffMember) => staffMember.name).join(", ")}
+        extra={
+          <Button type="primary" size={"middle"}>
+            Join Queue
+          </Button>
+        }
+      >
+        <h1>{queue.room}</h1>
+        {staffList.map((staffMember) => (
+          <div style={{ float: "left", padding: "0px 25px 0px 25px" }}>
```
dajinchu, line 34: i think we need to add a key
https://reactjs.org/docs/lists-and-keys.html

packages/app/pages/schedule.tsx
```diff
@@ -19,15 +23,18 @@ export default function Schedule() {
...
       end: e.endTime,
     })) ?? [];
-
+  // TODO: frontend guru pls cleanup height <3, month view is crunked :)
+  // esp on mobile
   return (
-    <div style={{ height: "400px" }}>
-      <h1>{data?.name ?? ""}</h1>
-      <Calendar localizer={momentLocalizer(moment)} events={myEvents} />
+    <div style={{ height: "70%" }}>
```
dajinchu, line 32: use styled-components?

packages/app/pages/today.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    <div>
+      <Navbar />
+      <Row>
+        <Col md={12} sm={24}>
+          {data?.map((q) => (
+            <QueueCard queue={q} />
+          ))}
+          {isTA ? (
+            <div style={{ padding: "0px 15px 0px 0px" }}>
```
dajinchu, line 68: styling here is kinda messed up. can we use Ant design grid stuff for positioning/spacing?

NEUDitao: https://ant.design/components/grid/#components-grid-demo-gutter looks promising

packages/app/pages/schedule.tsx
```diff
@@ -4,8 +4,12 @@ import useSWR from "swr";
 import { API } from "@template/api-client";
 import { Result } from "antd";
 
-export default function Schedule() {
-  const { data, error } = useSWR(`api/v1/courses/1`, async () =>
+type ScheduleProps = {
+  viewType: Calendar.Views; // TODO: debug this, hwy isn't it finding the types?
```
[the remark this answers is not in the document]
willstenzel: Cool, thanks!

packages/app/pages/schedule.tsx
```diff
@@ -19,15 +23,18 @@ export default function Schedule() {
...
+  // esp on mobile
   return (
-    <div style={{ height: "400px" }}>
-      <h1>{data?.name ?? ""}</h1>
-      <Calendar localizer={momentLocalizer(moment)} events={myEvents} />
+    <div style={{ height: "70%" }}>
+      <Calendar
+        localizer={momentLocalizer(moment)}
+        events={myEvents}
+        defaultView={viewType}
```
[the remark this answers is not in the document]
willstenzel: No problem. We got it fixed

packages/app/pages/schedule.tsx
```diff
@@ -4,8 +4,12 @@ import useSWR from "swr";
...
-export default function Schedule() {
-  const { data, error } = useSWR(`api/v1/courses/1`, async () =>
+type ScheduleProps = {
+  viewType: Calendar.Views; // TODO: debug this, hwy isn't it finding the types?
+};
+
+export default function Schedule({ viewType }: ScheduleProps) {
+  const { data, error } = useSWR(`api/v1/courses/1/schedule`, async () =>
     API.course.get(1)
   );
```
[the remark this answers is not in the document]
willstenzel: Good call on this one. We'll have it re-fetch the queue on focus, but not for the schedule

packages/common/index.ts
```diff
@@ -190,14 +210,20 @@ export enum QuestionType {
...
 // Ticket Status - Represents a given status of as student's ticket
-type QuestionStatus =
-  | "Drafting"
-  | "Queued"
-  | "Helping"
-  | "Resolved"
-  | "Deferred"
-  | "No Show"
-  | "Deleted";
+type QuestionStatus = OpenQuestionStatus | ClosedQuestionStatus;
```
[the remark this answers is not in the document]
NEUDitao: we should talk about this during the meeting tmrw 🦚

packages/app/pages/today.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  return (
+    <div>
+      <Navbar />
+      <Row gutter={25}>
+        <Col md={12} sm={24}>
+          {data?.map((q) => (
+            <QueueCard key={q.id} queue={q} />
+          ))}
+          {isTA ? (
```
liustanley, line 83: Minor change, but you can change this to `isTA && (<CreateQueueButton>...)`

packages/app/pages/today.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const CreateQueueButton = styled(Button)`
+  bottom: 0%;
+  width: 20%;
+  float: right;
+  background-color: #4cbb17;
+  color: white;
+`;
+
+const QueueCard = ({ queue }: QueueCardProps) => {
```
liustanley, line 38: Could this component be moved into the components folder?

packages/app/pages/schedule.tsx
```diff
@@ -19,15 +30,18 @@ export default function Schedule() {
 
   const myEvents =
     data?.officeHours.map((e) => ({
-      title: e.title,
       start: e.startTime,
       end: e.endTime,
     })) ?? [];
-
+  // TODO: frontend guru pls cleanup height <3, month view is crunked :)
+  // esp on mobile
```
liustanley: I found a solution for the month view, turns out your calendar needs to have a set height in order for it to display properly. I just pushed a commit with my change for this.

Regarding responsiveness, it seems like the crunkiness is coming from the Toolbar. It's probably worth looking into a custom Toolbar if you can't find a built in responsive solution in the library: https://jquense.github.io/react-big-calendar/examples/index.html#prop-components. This seems like a good example of a custom Toolbar: https://github.com/jquense/react-big-calendar/issues/191.

packages/app/pages/today.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const AvatarContainer = styled.div`
+  padding-left: 25px;
+  padding-right: 25px;
+  float: left;
+`;
+
+const CreateQueueButton = styled(Button)`
+  bottom: 0%;
+  width: 20%;
```
liustanley, line 32: We generally don't want to make widths percentages because it makes things really funky on small screen sizes. This can be changed to a fixed width or just not specified width is also fine.

packages/app/pages/schedule.tsx
```diff
@@ -19,15 +30,15 @@ export default function Schedule() {
...
   return (
-    <div style={{ height: "400px" }}>
-      <h1>{data?.name ?? ""}</h1>
-      <Calendar localizer={momentLocalizer(moment)} events={myEvents} />
-    </div>
+    <ScheduleCalendar
+      localizer={momentLocalizer(moment)}
+      events={myEvents}
+      defaultView={viewType}
+    />
```
liustanley, line 42: I found a solution for the month view, turns out your calendar needs to have a set height in order for it to display properly. I just pushed a commit with my change for this.

Regarding responsiveness, it seems like the crunkiness is coming from the Toolbar. It's probably worth looking into a custom Toolbar if you can't find a built in responsive solution in the library: https://jquense.github.io/react-big-calendar/examples/index.html#prop-components. This seems like a good example of a custom Toolbar: jquense/react-big-calendar#191.

packages/app/pages/today.tsx
```diff
@@ -0,0 +1,95 @@
...
+        title="Something went wrong, please ask chinese man"
+      />
+    );
+  }
+
+  return (
+    <div>
+      <Navbar />
+      <Row gutter={25}>
+        <Col md={12} sm={24}>
```
liustanley, line 79: Right now the card column acts weirdly in xs screen sizes, for both of these Cols I think all you have to do is change sm to xs since xs is the smallest breakpoint.

packages/app/pages/schedule.tsx
```diff
@@ -1,12 +1,23 @@
...
+
+type ScheduleProps = {
+  viewType: View;
+};
+
+export default function Schedule({ viewType }: ScheduleProps) {
+  const { data, error } = useSWR(
+    `api/v1/courses/1/schedule`,
+    async () => API.course.get(1),
+    { revalidateOnFocus: false }
```
dajinchu, line 20: why is this false

dajinchu: Think about people that leave 50 tabs open. It might be a week later when they switch back to the tab. I don't see any benefit to disabling revalidateOnFocus.

dajinchu: True though that the schedule doesn't change much. This just seems like premature optimization.

bugsalexander: Those are good points. I no longer feel strongly either way, so I'd say it's your call @willstenzel @NEUDitao

willstenzel: I agree with @dajinchu on this one. I'll change it to not disable `revalidateOnFocus`

PR #10, opened 2020-06-11
Add Login page, NavBar component and format

ofloody:


packages/app/pages/login.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const onSubmit = () => {
+    if (!username || !password) {
+      alert("Must input username and password!");
+    }
+  };
+
+  return (
+    <Container>
+      <LoginContainer>
+        <div>{username}</div>
```
willstenzel, line 53: Is there a reason the username is showing above the form input?

packages/app/components/Nav/NavBar.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const Container = styled.div`
+  width: 1440px;
+  height: 64px;
+`;
+interface NavBarProps {
+  course: Course;
+}
+
+export default function NavBar({ course }: NavBarProps) {
```
willstenzel, line 15: Looks like the course is coming in as props but is not used in the component. I think you'll want to use `course.name` where is says "Course" on line 29

packages/app/styles/NavBar.css
```diff
@@ -0,0 +1,109 @@
+@import "~antd/dist/antd.css";
+
```
willstenzel, line 2: Screen Shot 2020-06-11 at 5 38 46 PM
When I moved the screen to a smaller size, I saw that the items were stacking before the drawer button was shown. The [Ant design menu](https://ant.design/components/menu/) might be able to help with that.

packages/app/styles/NavBar.css
```diff
@@ -0,0 +1,109 @@
...
+.menuBar {
```
NEUDitao, line 3: we should use styled components

packages/app/styles/NavBar.css
```diff
@@ -0,0 +1,109 @@
...
+    margin-left: -20px;
+  }
+  .menuCon .ant-menu-item,
+  .menuCon .ant-menu-submenu-title {
+    padding: 1px 20px;
+  }
+  .logo a {
+    padding: 10px 20px;
+  }
+}
```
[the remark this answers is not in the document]
liustanley: Oh yeah that's my bad I can fix that, was using a component we found online.

packages/app/pages/login.tsx
```diff
@@ -0,0 +1,92 @@
...
+    setUsername(e.target.value);
+  };
+
+  const onChangePassword = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setPassword(e.target.value);
+  };
+
+  const onSubmit = () => {
+    if (!username || !password) {
+      alert("Must input username and password!");
```
aislinblack, line 46: Does it make sense to both alert the user and show it in the input? I think it's mostly just added fuss to have to click out of the alert

packages/app/styles/NavBar.css
ofloody: Stanley and I agreed not to use the Ant design menu because it wasn't reactive. That's way there is a css file. I will try to fix the spacing when the screen is in that awkward size.

packages/app/components/Nav/NavBar.tsx
ofloody: I left this unfinished because I thought that there would need to be a drop down added to the course name, so when you clicked on it the list of your courses would appear. There wasn't a course data definition in there when I did the pr so I couldn't use the field. I can go back and change it now that I have the version with the data definitions.

packages/app/pages/login.tsx
ofloody: I think the alert was just to stop someone from submitting if they haven't filled in both, but when we are actually able to verify if they input a valid username and password then I agree we wouldn't need that alert.

packages/app/styles/NavBar.css
willstenzel: Sounds good

packages/app/components/Nav/LeftNavBar.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { Component } from "react";
+import { Menu } from "antd";
+const SubMenu = Menu.SubMenu;
+const MenuItemGroup = Menu.ItemGroup;
+
+interface LeftNavBarProps {}
+
+export default function LeftNavBar({}: LeftNavBarProps) {
```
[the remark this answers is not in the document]
ofloody: Good point

packages/app/components/Nav/NavBar.tsx
ofloody: I think this should probably be done in a separate PR, so I'm going to leave it for now?

willstenzel: Yeah that's all good

PR #11, opened 2020-06-11
TA Queue UI

liustanley:
Refactored student and TA queues into the same components, in `/pages/queue/index.tsx`, change `ROLE` in order to toggle between the two.


packages/app/components/Queue/GroupQuestions.tsx
```diff
@@ -0,0 +1,78 @@
...
+const AddButton = styled(Button)`
+  margin-top: 8px;
+`;
+
+const CenterRow = styled(Row)`
+  align-items: center;
+  padding-right: 4px;
+`;
+
+const GroupQuestions = () => {
```
NEUDitao, line 30: can you render these with mock data? an example could be `/mocks/listQuestions.ts`. You don't have to set up an endpoint or anything, but it'd be nice to have the code here to process that mock data so eventually when we set up the endpoint it requries no work

packages/app/components/Queue/QueueCard.tsx
```diff
@@ -31,47 +54,121 @@ const CenterRow = styled(Row)`
...
-        <Col span={1}>
-          <Text>{rank}</Text>
-        </Col>
-        <Col xs={16} sm={11} lg={6}>
-          <CenterRow>
-            <Avatar icon={<UserOutlined />} />
-            <NameContainer>
-              <Text>{name}</Text>
-            </NameContainer>
+    <div>
```
NEUDitao, line 78: is it possible to either:
- split this up into three separate things so they're more readable
- localize the changes? I know there are quite a few values in here that're tricky to convey (such as some cards having xxl widths and some not), but I feel like that'd be nicer to read than this, but idk if that's just me

packages/app/components/Queue/QueueList.tsx
```diff
@@ -42,75 +52,186 @@ const CenterRow = styled(Row)`
...
+interface QueueListProps {
+  role: Role;
+  onOpenClick: (name: string) => void;
+}
+
+export default function QueueList({ role, onOpenClick }: QueueListProps) {
+  const [helping, setHelping] = useState<boolean>(true);
+  const screens = useBreakpoint();
+
+  console.log(screens);
```
NEUDitao, line 76: was this for debugging?

packages/app/components/Queue/QueueList.tsx
```diff
@@ -42,75 +52,186 @@ const CenterRow = styled(Row)`
...
+          )}
 
-      <HeaderCard bordered={false}>
-        <CenterRow justify="space-between">
-          <Col span={1}>
-            <HeaderText>#</HeaderText>
-          </Col>
-          <Col xs={16} sm={11} lg={6}>
-            <HeaderText>name</HeaderText>
+          <QueueCard
```
NEUDitao, line 166: same as above

packages/app/components/Queue/QueueCard.tsx
```diff
@@ -31,47 +54,121 @@ const CenterRow = styled(Row)`
...
+      {role === "ta" && !helping && (
```
NEUDitao, line 79: for the `"ta"` and `"student"`, could we make `Role` an `enum` instead of union string to have more consistent typing?

packages/app/components/Queue/QueueCard.tsx
liustanley: yeah for sure! i think that'd make things more readable. i'm not sure exactly what you're suggesting, do you think separating each case into three different render methods would be sufficient?

packages/app/components/Queue/GroupQuestions.tsx
liustanley: yeah for sure! this PR was only intended for the UI though, we're planning on loading mock data for this week's sprint so I'll leave that task for our next PR

packages/app/components/Queue/QueueCard.tsx
liustanley: i just pushed my changes that address this but i could've misunderstood your request, just reopen this issue if more can be done!

packages/app/components/Queue/StudentInfoCard.tsx
```diff
@@ -0,0 +1,86 @@
...
+const StudentInfoCard = () => {
+  return useMemo(() => {
+    return (
+      <HelpCard
+        headStyle={{ padding: "0 16px" }}
+        bodyStyle={{ padding: "16px" }}
+        title={
+          <Row justify="space-between">
+            <Name>Alex Takayama</Name>
+            <Email>takayama.a@northeastern.edu</Email>
```
willstenzel, line 56: Is the email necessary here? If not, it might work to put the finish helping button there to save some space

packages/app/components/Queue/GroupQuestions.tsx
```diff
@@ -0,0 +1,78 @@
...
+  return useMemo(() => {
+    return (
+      <Collapse>
```
willstenzel, line 33: Just curious, would we ever need a max hight to be set on the `Collapse` component?

packages/app/components/Queue/StudentInfoCard.tsx
liustanley: Not sure if the email is necessary, but I like being able to see the student's email on waitwhile so that I can make sure I'm contacting the right student. I think it might be easier for the user if all the buttons are in one place, I tried putting the finish button in the header before and it felt out of place.

packages/app/components/Queue/GroupQuestions.tsx
liustanley: The height depends on how many questions there are in the queue, but it might be cool to make it a scrollable list of questions to save space. I'm not sure which UI is preferable, but I played around a bit with css and it's a little weird trying to hide the scrollbars within ant design. Maybe this could be pushed off for later?

packages/app/components/Queue/StudentInfoCard.tsx
willstenzel: That makes a lot of sense. We should leave the email as it is then

packages/app/components/Queue/GroupQuestions.tsx
willstenzel: Yeah a scrollable list would be cool but it’s definitely not a top priority right now so we can revisit it later

Issue #1, opened 2020-04-20
Create API Docs


willstenzel:

willstenzel:
Don't think we need this anymore because the api-client describes all the operations that can be done as well as the input and response types for the data.

PR #14, opened 2020-06-14
Add TypeORM Entities and Fixtures plus update some endpoints

willstenzel:
The focus of this PR was creating all the TypeORM Entities and Fixtures. 
For those unfamiliar, TypeORM entities define the models for the database and the Fixtures populate those models with seed data. 

We also changed the `courses/{course_id}/queues` and the `queues/{queue_id}/questions` endpoints to use real db data instead of mock jsons. There is more work to be done to convert all the endpoints to be use the db data so this part of the PR but that can be tackled in a future PR.


packages/server/src/testUtils.ts
```diff
@@ -60,3 +66,33 @@ export function withServer(server) {
     expect(request.result).toStrictEqual(result);
   };
 }
+
+export const generateMockData = async () => {
```
dajinchu, line 70: This abstraction is sure to bite us in the butt later. How do we extend the mock data without affecting every other test? How do we remember all the different entities being mocked? How do we avoid mocking 4 other things that aren't actually needed for a specific test?

Two proposals:
1. Just create data manually in every single test. This is verbose, but it makes it obvious to the reader what data exists without hopping between files. It also makes it easy to have different data for different tests.
2. Make small, descriptive mock helpers. `const FUNDIES = async()=>CourseModel.create({name:"CS 2500", icalUrl: "fundies.com"})`

we can do both of the above, but in the name of Avoiding Hasty Abstractions, maybe do 1 before 2.

packages/server/src/api/courseRoutes.ts
```diff
@@ -41,24 +42,17 @@ export const courseRoutes: ServerRoute[] = [
     method: "GET",
     path: "/api/v1/courses/{course_id}/queues",
     handler: async (request, h): Promise<GetCourseQueuesResponse> => {
-      const queuesResponse = cloneDeep(MOCK_GET_COURSE_RESPONSE.queues);
+      const queues = await QueueModel.find({
+        where: { course_id: request.params.course_id },
```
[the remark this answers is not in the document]
willstenzel: Yep, we just added a TODO

packages/server/src/api/courseRoutes.ts
```diff
@@ -41,24 +42,17 @@ export const courseRoutes: ServerRoute[] = [
...
+      });
 
-      queuesResponse.forEach(
-        (queue) =>
-          (queue["queueSize"] = queue.questions.filter(
-            (question) => question.status in OpenQuestionStatus
-          ).length)
-      );
+      for (let queue of queues) {
+        queue["queueSize"] = (await queue.questions).length;
```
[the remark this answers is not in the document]
willstenzel: That `count` definitely cleans things up. Thanks!

packages/server/src/api/courseRoutes.ts
```diff
@@ -41,24 +42,17 @@ export const courseRoutes: ServerRoute[] = [
...
+        queue["staffList"] = MOCK_GET_COURSE_RESPONSE.queues[0].staffList;
```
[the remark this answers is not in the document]
willstenzel: Good call

packages/server/fixtures/semesterModel.yml
```diff
@@ -0,0 +1,5 @@
+entity: SemesterModel
```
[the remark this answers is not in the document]
willstenzel: Thought we added it. Should be in `packages/server/src/entity/SemesterModel.ts`

packages/server/src/api/queueRoutes.ts
```diff
@@ -27,9 +31,43 @@ export const queueRoutes: ServerRoute[] = [
...
+            "status",
+            "name",
+            "text",
+            "taHelped",
+            "helpedAt",
+            "closedAt",
+            "questionType",
+          ]);
+          result.taHelped = await result.taHelped;
+          result.creator = pick(await (await result.creator).user, [
```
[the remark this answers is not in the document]
willstenzel: Yeah, I was thinking it felt kind of odd to but I guess we'll have to roll with it.

packages/server/src/testUtils.ts
willstenzel: That makes sense, I went with #1 for now and we can revisit the abstraction later.

packages/server/src/api/queueRoutes.ts
```diff
@@ -137,3 +118,43 @@ export const queueRoutes: ServerRoute[] = [
     },
   },
 ];
+
+// for some reason, JOI.allow(null) means the property has to exist, but can be null
```
bugsalexander, line 122: > for some reason, `property: JOI.allow(null)` means the property has to exist, but can be null

I would have thought that the two would be synonymous. this is annoying. thoughts?

packages/server/src/api/queueRoutes.ts
```diff
@@ -137,3 +118,43 @@ export const queueRoutes: ServerRoute[] = [
...
+
+async function questionModelToQuestion(qm: QuestionModel): Promise<Question> {
+  return {
+    creator: await userModelToUserPartial(await (await qm.creator).user),
+    id: qm.id,
+    createdAt: qm.createdAt,
+    status: qm.status,
+    text: qm.text,
+    // qm.taHelped: types says is nonnullable, but it is nullable
```
bugsalexander, line 131: > qm.taHelped: typescript says it is nonnullable, but it is nullable

TS says property exists with correct type (nonnullable), but it is nullable (for example, run the tests). this is a mistake with the types we have currently defined, imo, because it makes sense that `taHelped` might not exist.

packages/server/src/api/queueRoutes.ts
```diff
@@ -137,3 +118,43 @@ export const queueRoutes: ServerRoute[] = [
...
+    taHelped:
+      (await qm.taHelped) &&
+      (await userCourseModelToUserPartial(await qm.taHelped)),
```
bugsalexander, line 134: if property `taHelped` exists on the QuestionModel, then convert it to a UserPartial. if not, then leave it as null.

packages/server/src/api/queueRoutes.ts
bugsalexander: imo this can be fixed in a separate PR. I think it's fine for now as long as we remember this weirdness

packages/server/src/api/queueRoutes.ts
bugsalexander: imo this should be changed asap. seems like a mistake.

packages/server/src/api/queueRoutes.ts
bugsalexander: this is needed because of the above (taHelped may be null, in which case don't try to convert)

packages/common/index.ts
```diff
@@ -230,12 +230,15 @@ export enum OpenQuestionStatus {
...
-  NoShow = "No Show",
+  NoShow = "NoShow",
   Deleted = "Deleted",
 }
 
 // Ticket Status - Represents a given status of as student's ticket
-export const QuestionStatus = {
+export type QuestionStatus = keyof typeof QuestionStatusKeys;
+// IMPORTANT: whenever importing QuestionStatusKeys to use for any reason, import it from
+// @template/common/index. for some reason, importing from @template/common results in undefined.
```
bugsalexander, line 240: this bug is absolutely disgusting. took me an hour before I was able to find out this issue exists.

willstenzel: Ouch that looks like a tough one. Nice work hunting down the issue!

packages/server/src/api/queueRoutes.ts
bugsalexander: update: will and I talked about this, and it is already nullable, typescript just is weird.

packages/server/fixtures/semesterModel.yml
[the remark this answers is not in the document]
dajinchu: yep you right

package.json
```diff
@@ -6,10 +6,11 @@
     "dev": "concurrently -n \"common,server,app,proxy\" \"yarn workspace @template/common dev\" \"yarn workspace @template/server dev\" \"yarn workspace @template/app dev\" \"yarn dev:proxy\"",
     "dev:proxy": "node infrastructure/dev/devProxy.js",
     "tsc": "yarn workspaces run tsc",
-    "test": "jest",
+    "test": "echo 'CREATE DATABASE test;' | psql -U postgres -h localhost && jest",
     "dev:db:up": "docker-compose -f infrastructure/dev/docker-compose.yml up -d && yarn dev:db:seed",
     "dev:db:down": "docker-compose -f infrastructure/dev/docker-compose.yml down",
     "dev:db:seed": "yarn workspace @template/server dev:db:seed",
+    "dev:db:reset": "echo 'DROP DATABASE dev;' | psql -U postgres -h localhost && echo 'CREATE DATABASE dev;' | psql -U postgres -h localhost",
```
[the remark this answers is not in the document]
dajinchu: for test scripts I believe we should use a `test:db` prefix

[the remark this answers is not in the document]
NEUDitao: test database is always basically empty anyways cause the data is created per test, so it should always be fine :)

PR #16, opened 2020-06-16
Dynamic queue using mock endpoints

aislinblack:
Modified Queue UI to use context to keep track of profile. Get questions from the api.


packages/app/components/Queue/QueueList.tsx
```diff
@@ -170,91 +199,63 @@ export default function QueueList({ role, onOpenClick }: QueueListProps) {
...
-            )}
-          </Row>
-          {role === "ta" && !helping && renderTAHeader()}
-          {role === "ta" && helping && renderHelpingHeader()}
-          {role === "student" && renderStudentHeader()}
+              )}
+            </Row>
+            {role === Role.TA && !helping && renderTAHeader()}
+            {role === Role.TA && helping && renderHelpingHeader()}
+            {role === Role.STUDENT && renderStudentHeader()}
```
willstenzel, line 239: This totally works but I'm curious if it would be easier long term to have a single if statement on the `role` attribute that could render both the TA Header and the QueueCard in the same file instead of separating the logic into different files? This would mean making a separate card for the student and ta (`StudentQueueCard` and `TAQueueCard`) so I'm not sure it's worth the refactoring effort at the moment but just thought I'd throw it out there.

packages/app/pages/queue.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+export default function Queue({}: QueueProps) {
+  const [isJoining, setIsJoining] = useState<boolean>(false);
+  const [openPopup, setOpenPopup] = useState<boolean>(false);
+  const { profile } = useContext(ProfileContext);
+  const [course, setCourse] = useState<UserCourse>(null);
+  const [questions, setQuestions] = useState<Question[]>([]);
+  const [queueId, setQueueId] = useState<number>(null);
+  const [questionDraftId, setQuestionDraftId] = useState<number>(null);
+  const [currentQuestion, setCurrentQuestion] = useState<Question>(null);
```
willstenzel, line 48: There's a lot going on in this component which is not necessarily a bad thing but it can make it hard to easily understand what is going on. Not an issue for this PR but in the future, if we add more complexity, it might be worth considering separating this out into a `TAQueue` component and a `StudentQueue` component which could be rendered from a conditional in the parent `Queue` component.

packages/app/pages/queue.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * Sends a push notification to the student with the given Question
+   * @param question the question to be notified
+   */
+  const alertStudent = (question: Question) => {
+    // Send API request to trigger notification
+  };
+
+  return useMemo(() => {
```
willstenzel, line 160: Good call on the `useMemo` here

packages/app/pages/queue.tsx
aislinblack: I think it definitely makes sense to separate this out into two components. Perhaps, it would make a good smaller ticket for the future?

packages/app/pages/queue.tsx
aislinblack: Honestly I don't know its mostly habit from work where we memoize everything to reduce re-rendering.

liustanley: Removed the useMemo calls for now, we agreed not to use them unless we have a noticeable performance drop

packages/app/components/Queue/QueueList.tsx
liustanley: I'm not sure if I completely understand, but the headers and queue card rendering logic has to be separate because there's only one header for a list of queue cards. I found that there were a lot of similarities between the StudentQueueList/StudentQueueCard and the TAQueueList/TAQueueCard so I abstracted them into the same files. I'm not sure if this detracts from readability though, lmk if you have any further thoughts on this.

willstenzel: That makes sense. Looking back to my original comment I wasn't so clear. 
What I was thinking was that it could more readable to take each function in the `QueueCard` like `renderTACard()`, `renderHelpingCard()`, `renderStudentCard()` and separate them into their own components. Then in the `QueueList` you could have something like 
```
{if (role === Role.TA && !helping) {
  renderTAHeader();
  {questions.map((question: Question, index: number) => {
     return (<TAQueueCard />)   // which would be the same as the renderTACard() function
  }
}
```
It would add some more code to the `QueueList` component so I'm not sure it would be that much better. I'm curious to hear what people think.

packages/app/components/Queue/GroupQuestions.tsx
```diff
@@ -1,7 +1,9 @@
-import React, { useMemo } from "react";
+import React, { useMemo, useState, Fragment } from "react";
```
bugsalexander, line 1: useMemo and Fragment are unused imports. i'll remove them for you

packages/app/components/Queue/QueueList.tsx
```diff
@@ -191,67 +252,29 @@ export default function QueueList({ role, onOpenClick }: QueueListProps) {
...
-            onOpen={onOpenClick}
-          />
+          {questions.map((question: Question, index: number) => {
+            const creator = question.creator;
+            return (
+              <QueueCard
+                key={question.id}
+                helping={helping}
+                role={role}
+                rank={99}
```
bugsalexander, line 271: should this be `rank={index}`?

packages/app/pages/queue.tsx
```diff
@@ -0,0 +1,234 @@
...
+      getQuestions();
+    }
+  }, [queueId]);
+
+  /**
+   * Gets the questions for this course
+   */
+  const getQuestions = () => {
+    if (queueId) {
+      API.questions.index(queueId).then((q: Question[]) => {
```
bugsalexander, line 73: can use async/await for less code:
```
const getQuestions = async () => {
  if (queueId) {
    const q = await API.questions.index(queueId);
    if (q) {
    ...
```

also, is it possible to use useSWR for this?

packages/app/components/Queue/StudentPopupCard.tsx
```diff
@@ -69,80 +74,84 @@ const BodyText = styled.div`
...
+        <ButtonDiv>
+          <RemoveButton
+            danger
+            block
+            onClick={() =>
+              updateQuestion(question, ClosedQuestionStatus.Deleted)
+            }
+          >
+            Remove from Queue
+          </RemoveButton>
```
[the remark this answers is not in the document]
liustanley: Good catch, the state wasn't properly updating upon removal just fixed it.

packages/app/pages/queue.tsx
liustanley: We were having trouble figuring out how to use SWR for this because we have to wait until the `/profile` request is fulfilled from the `ProfileContext` in order to send a request for the questions. I can open an issue for this if we can't find a quick solution?

packages/app/components/Queue/QueueList.tsx
liustanley: Oh I see! I think that's valid and might be more readable. I'm inclined to keep the `QueueCards` in the same component for now though so that we can reuse the same styled-components. I also think it might not be worth the refactoring effort at the moment, but I'd be happy to revisit this if it poses issues later on.

packages/app/pages/queue.tsx
liustanley: Referenced here: #19

PR #23, opened 2020-06-20
Factories + un-lazy models

dajinchu:


packages/server/src/api/queueRoutes.test.ts
```diff
@@ -15,55 +17,29 @@ describe("Queue Routes", () => {
...
+          helpedAt: null,
+          id: 1,
+          questionType: "Other",
+          status: "Queued",
+          taHelped: null,
+          text: "Help pls",
+        },
+      ]);
     });
     // TODO: is this test supposed to fail now?
```
bugsalexander, line 44: can you get rid of this TODO stub?

packages/server/src/api/queueRoutes.test.ts
```diff
@@ -75,16 +51,25 @@ describe("Queue Routes", () => {
...
+      expect(request.statusCode).toEqual(201);
+      expect(request.result).toMatchObject({
+        text: "Don't know recursion",
+        taHelped: null,
+        helpedAt: null,
+        closedAt: null,
+        questionType: "Concept",
+        status: "Queued",
+      });
+      expect(QuestionModel.count({ where: { queueId: 1 } })).toEqual(1);
```
bugsalexander, line 72: could you duplicate this line to right before the POST? except with expected count as `0` instead.

packages/server/src/api/queueRoutes.ts
```diff
@@ -122,25 +122,23 @@ export const queueRoutes: ServerRoute[] = [
...
+    creator: userModelToUserPartial(qm.creator),
     id: qm.id,
     createdAt: qm.createdAt,
     status: parseStatus(qm.status),
     text: qm.text,
     // qm.taHelped: types says is nonnullable, but it is nullable
-    taHelped:
-      (await qm.taHelped) &&
-      (await userCourseModelToUserPartial(await qm.taHelped)),
+    taHelped: qm.taHelped && userModelToUserPartial(qm.taHelped),
```
bugsalexander, line 133: Does this typecheck? When I wrote this code, this errored, because qm.taHelped wasn't a `UserModel`.

packages/server/src/entity/QuestionModel.ts
```diff
@@ -17,29 +19,29 @@ export class QuestionModel extends BaseEntity {
...
   queueId: number;
 
   @Column("text")
   text: string;
 
-  @ManyToOne((type) => UserCourseModel)
+  @ManyToOne((type) => UserModel)
   @JoinColumn({ name: "creatorId" })
-  creator: Promise<UserCourseModel>;
+  creator: UserModel;
```
bugsalexander, line 32: Ah, disregard previous comment, as I see that now the types should line up with this change.

Was this a mistake we had made earlier?

willstenzel:
@bugsalexander thanks for help cleaning things up!

packages/server/src/entity/QuestionModel.ts
dajinchu: yeah it wasn't exactly a mistake but not optimal to be using the UserCourseModel

packages/server/src/api/queueRoutes.test.ts
willstenzel: Done.

packages/server/src/api/queueRoutes.test.ts
willstenzel: Good call, just added that in.

dajinchu:
> 
> I am still getting test failures – do they pass on your machines?

I'm seeing them pass, except for queueRoutes returning 200 instead 201 status code which is fine

willstenzel:
> I am still getting test failures – do they pass on your machines?
Same as @dajinchu on my machine too

PR #24, opened 2020-06-22
Github Actions

dajinchu:
Create Github Actions to ensure linting and run tests. Runs on pull requests.

Also makes server tests run-in-band by default so you don't have to pass the flag

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "dev": "concurrently -n \"common,server,app,proxy\" \"yarn workspace @template/common dev\" \"yarn workspace @template/server dev\" \"yarn workspace @template/app dev\" \"yarn dev:proxy\"",
     "dev:proxy": "node infrastructure/dev/devProxy.js",
     "tsc": "yarn workspaces run tsc",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
     "dev": "concurrently -n \"common,server,app,proxy\" \"yarn workspace @template/common dev\" \"yarn workspace @template/server dev\" \"yarn workspace @template/app dev\" \"yarn dev:proxy\"",
     "dev:proxy": "node infrastructure/dev/devProxy.js",
     "tsc": "yarn workspaces run tsc",
-    "test": "jest",
+    "test": "yarn --runInBand",
```
bugsalexander, line 9: should this be `jest --runInBand`?

willstenzel:
Thanks for setting this up. I just had two quick questions:
1. Will the lint check always pass if we have husky lint already set to run when we commit?
2. Would it make sense to set up the [github super linter](https://github.com/github/super-linter) to get some cleanup on our non-js files as well?

PR #34, opened 2020-06-24
Add support for text notifications with Twilio

bugsalexander:


packages/server/src/api/notifRoutes.test.ts
```diff
@@ -63,3 +64,76 @@ describe("/api/v1/notifications/register/{user_id}", () => {
...
+      userId: 1,
+    });
+  });
+});
+
+describe("/api/v1/notifications/notify_user/{user_id}", () => {
+  setupDBTest();
+  const getServer = setupServerTest();
+
+  it("tests that invalid endpoints for web push get deleted in db", async () => {
```
willstenzel, line 105: Just curious, where in the code are the endpoints getting deleted?

NEUDitao: https://github.com/sandboxnu/office-hours/blob/eb4810c226c6c12bf99de5633bb82da393446213/packages/server/src/api/notifRoutes.ts#L135

PR #37, opened 2020-06-25
useProfile hook with redirect

liustanley:
Adds a custom hook, `useProfile`, that sends a request to `/api/v1/profile` to determine if the client is logged in or not. If the client is not logged in, the hook automatically redirects the user to `/login`.

I'm guessing that later we'll be storing the user authentication token in a cookie, so when that gets set up this hook can be modified to redirect users when this token doesn't exist on the frontend.


packages/app/pages/schedule.tsx
```diff
@@ -32,11 +34,15 @@ export default function Schedule({ viewType }: ScheduleProps) {
...
     })) ?? [];
 
-  return (
-    <ScheduleCalendar
-      localizer={momentLocalizer(moment)}
-      events={myEvents}
-      defaultView={viewType}
-    />
-  );
+  if (profile) {
```
dajinchu, line 37: I don't think we need to check for profile to exist since it'll just redirect? If we setup this properly to switch the view based on `data` and `error` that'll cover it, since non-logged in people will get error from the 401 unauthorized.

dajinchu: Same thing goes for `today` and `queue` of course.

dajinchu: See how they do the three cases (data, no data yet, error)
```
function Profile() {
  const { data, error } = useSWR('/api/user', fetcher)

  if (error) return <div>failed to load</div>
  if (!data) return <div>loading...</div>
  return <div>hello {data.name}!</div>
}
```
from https://swr.now.sh/#basic-data-loading

liustanley: Oh it's already set up to redirect for people who aren't logged in, the reason why I added that check is to prevent the `schedule` page from loading partially before redirecting. Then we can replace the return null with the loading shell UI once we finish that. I'm not sure if there's a better way to set this up?

liustanley: Like right now without the check, the `schedule` page will fully load before the `useProfile` hook fetches and redirects, so logged out users will also be able to see the page.

dajinchu: right, but once we merge this with the backend, when you are logged out, the actual schedule data will 401. Which, if we handle those cases properly, will show an error state to the user before redirecting, which is exactly what we want.

dajinchu: Point is, just handle the SWR data, error and we won't need to special case the `profile`

NEUDitao: > will show an error state to the user before redirecting

i dunno if we want that though -- most websites don't error and then send you to the login page, they just send you there ya feel?

liustanley: yeah that's what i was going for in my PR, i figured that either way we would have to have conditional rendering on `data` and `error` which is why i opted to only conditionally render on `profile`. i might be misunderstanding Da-Jin though, i'm not sure how else we would handle the SWR data error?

dajinchu: Gotcha. Let's leave it as is for now. There might be issues with static site generation in the future, but let's wait for these pages to be fully built to worry about those optimizations.

PR #43, opened 2020-06-27
Course and queue URL parameters

liustanley:
Adds navbar to pages and changes URL scheme to the following:

Today: `/class/{classId}/today`
Schedule: `/class/{classId}/schedule`
Queue: `/class/{classId}/queue/{queueId}`

### DIFF
--- a/packages/app/components/Nav/LeftNavBar.tsx
+++ b/packages/app/components/Nav/LeftNavBar.tsx
@@ -3,6 +3,7 @@ import styled from "styled-components";
 import { Menu } from "antd";
 const SubMenu = Menu.SubMenu;
 const MenuItemGroup = Menu.ItemGroup;
+import Link from "next/link";
 
 const HorizontalMenu = styled(Menu)`
   border-bottom: none;
@@ -18,23 +19,33 @@ const MenuItem = styled(Menu.Item)`
   }
 `;
 
-const MenuItemLink = styled.a`
+const MenuItemLink = styled(Link)`
   padding: 10px 15px;
 `;
 
+interface LeftNavBarProps {
+  courseId: number;
+  queueId: number;
+}
 
-export default function LeftNavBar() {
+export default function LeftNavBar({ courseId, queueId }: LeftNavBarProps) {
   return (
     <HorizontalMenu mode="horizontal">
       <MenuItem key="today">
-        <MenuItemLink href="/today">Today</MenuItemLink>
+        <MenuItemLink href={`/class/${courseId}/today`}>Today</MenuItemLink>
       </MenuItem>
       <MenuItem key="week">
-        <MenuItemLink href="/schedule">Schedule</MenuItemLink>
+        <MenuItemLink href={`/class/${courseId}/schedule`}>
+          Schedule
+        </MenuItemLink>
       </MenuItem>
-      <MenuItem key="queue">
-        <MenuItemLink href="/queue">Queue</MenuItemLink>
-      </MenuItem>
+      {queueId && (
+        <MenuItem key="queue">
+          <MenuItemLink href={`/class/${courseId}/queue/${queueId}`}>
+            Queue
+          </MenuItemLink>
+        </MenuItem>
+      )}
     </HorizontalMenu>
   );
 }
--- a/packages/app/components/Nav/LeftNavBar.tsx
+++ b/packages/app/components/Nav/LeftNavBar.tsx
@@ -19,10 +19,6 @@ const MenuItem = styled(Menu.Item)`
   }
 `;
 
-const MenuItemLink = styled(Link)`
-  padding: 10px 15px;
-`;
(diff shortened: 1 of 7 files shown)


packages/app/components/Nav/LeftNavBar.tsx
```diff
@@ -18,23 +19,33 @@ const MenuItem = styled(Menu.Item)`
...
+  queueId: number;
+}
 
-export default function LeftNavBar() {
+export default function LeftNavBar({ courseId, queueId }: LeftNavBarProps) {
   return (
     <HorizontalMenu mode="horizontal">
       <MenuItem key="today">
-        <MenuItemLink href="/today">Today</MenuItemLink>
+        <MenuItemLink href={`/class/${courseId}/today`}>Today</MenuItemLink>
```
dajinchu, line 35: Good catch on using next/link, but we need to use the `as` prop or it won't be able to do link preloading.
https://nextjs.org/docs/api-reference/next/link#dynamic-routes
note: "href is a file system path used by the page and it shouldn't change at runtime"

packages/app/components/Nav/NavBar.tsx
```diff
@@ -98,11 +102,26 @@ const BarsButton = styled.span`
 `;
 
 interface NavBarProps {
-  course: Course;
+  profile: User;
+  courseId: number;
 }
 
-export default function NavBar({ course }: NavBarProps) {
+export default function NavBar({ profile, courseId }: NavBarProps) {
```
dajinchu, line 109: instead of passing profile through, I think we can just use `useProfile` hook

packages/app/components/Nav/LeftNavBar.tsx
```diff
@@ -18,23 +19,33 @@ const MenuItem = styled(Menu.Item)`
...
       <MenuItem key="week">
-        <MenuItemLink href="/schedule">Schedule</MenuItemLink>
-      </MenuItem>
-      <MenuItem key="queue">
-        <MenuItemLink href="/queue">Queue</MenuItemLink>
+        <MenuItemLink href={`/class/${courseId}/schedule`}>
+          Schedule
+        </MenuItemLink>
       </MenuItem>
+      {queueId && (
```
NEUDitao, line 42: should we remove the queue link if there's no queue? I think it should be a thing that's like "oh you need to sign up for a queue if you click on a queue that you don't have ya feel?"

packages/app/components/Nav/NavBar.tsx
```diff
@@ -98,11 +102,26 @@ const BarsButton = styled.span`
...
+  const [course, setCourse] = useState<CoursePartial>(
+    profile.courses.find((c) => c.course.id === courseId).course
+  );
+  const [queueId, setQueueId] = useState<number>(undefined);
+
+  const { data, error } = useSWR(`api/v1/courses/${courseId}/queue`, async () =>
+    API.course.queues(courseId)
+  );
+
+  useEffect(() => {
```
dajinchu, line 120: we don't need an effect for this. effects that set state will trigger extraneous re-renders.
just do:
```
const queueId = data && data.length > 0 && data[0].id
```

You can think of `queueId` as "derived" from `data`. It is not itself a piece of the state, it is derived from the `data` state.

packages/app/pages/index.tsx
```diff
@@ -7,7 +7,7 @@ export default function Home() {
   const profile: User = useProfile();
 
   if (profile) {
-    Router.push("/today");
+    Router.push("/class/" + profile.courses[0].course.id + "/today");
```
dajinchu, line 10: the rule for `next/link` applies for Router.push as well:
https://nextjs.org/docs/api-reference/next/router#usage
see
```
Router.push('/post/[pid]', '/post/abc')
```
first argument is the file system path to the page and should be static.

packages/app/components/Nav/LeftNavBar.tsx
liustanley: oh so the queue link actually navigates the user to the queue page of the open queue (if one exists), but if there's no queue then i think the link serves no purpose?

packages/app/components/Nav/LeftNavBar.tsx
liustanley: ohh awesome ok!

packages/app/components/Nav/NavBar.tsx
liustanley: good catch!

packages/app/components/Nav/NavBar.tsx
liustanley: got it, that makes sense. if `data` were to change, would `queueId` also update?

dajinchu: yeah, remember that the entire function component has to execute every time we re-render

PR #48, opened 2020-07-01
Created the ability to add staff notes to queues on the today page

NEUDitao:
image

us epic gamers even made it check if you're a TA and everything. sugondese!


packages/app/components/Today/OpenQueueCard.tsx
```diff
@@ -28,6 +37,53 @@ const OpenQueueCard = ({ queue }: OpenQueueCard) => {
...
+          ) : (
+            <p>{queue.notes}</p>
+          )}
+        </div>
+      )}
+      {!queue.notes &&
+        (editingNotes ? (
+          <Input
+            defaultValue={queue.notes}
+            onPressEnter={(value) => {
```
liustanley, line 71: these two `onPressEnter` functions are the same abstract them fundies dawg

packages/app/components/Today/OpenQueueCard.tsx
```diff
@@ -28,6 +37,53 @@ const OpenQueueCard = ({ queue }: OpenQueueCard) => {
...
+            />
+          ) : (
+            <p>{queue.notes}</p>
+          )}
+        </div>
+      )}
+      {!queue.notes &&
+        (editingNotes ? (
+          <Input
+            defaultValue={queue.notes}
```
liustanley, line 70: since `queue.notes` is an optional and this is the case where `queue.notes` is undefined, `defaultValue` could be left out or just made blank

packages/server/src/api/queueRoutes.test.ts
```diff
@@ -99,6 +104,29 @@ describe("Queue Routes", () => {
...
+      const user = await UserFactory.create();
+      const question = await QuestionFactory.create({
+        text: "Recursion is wrecking me",
+      });
+
+      const request = await injectAsUser(getServer(), user, {
+        method: "patch",
+        url: `/api/v1/queues/${question.queueId}`,
+        payload: {
+          notes: "dajin really loves sucking nuts from sugondese",
```
liustanley, line 118: rip

packages/app/components/Today/OpenQueueCard.tsx
NEUDitao: have you ever t yped copy paste

packages/app/components/Today/OpenQueueCard.tsx
```diff
@@ -16,8 +20,23 @@ const AvatarContainer = styled.div`
...
+  float: right;
+`;
+
+const OpenQueueCard = ({ queue, isTA }: OpenQueueCard) => {
+  const [editingNotes, setEditingNotes] = useState(false);
   const staffList = queue.staffList;
+
+  const updateNotes = (value: any) => {
+    const notes = (value.target as any).value;
+    if (notes) {
```
dajinchu, line 33: this should probably check if `notes !== queue.notes`

packages/app/components/Today/OpenQueueCard.tsx
```diff
@@ -16,8 +20,23 @@ const AvatarContainer = styled.div`
...
+const EditOutlinedFloatedRight = styled(EditOutlined)`
+  float: right;
+`;
+
+const OpenQueueCard = ({ queue, isTA }: OpenQueueCard) => {
+  const [editingNotes, setEditingNotes] = useState(false);
   const staffList = queue.staffList;
+
+  const updateNotes = (value: any) => {
+    const notes = (value.target as any).value;
```
dajinchu, line 32: we probably wanna make it a controlled component because right now if you put in a note, and then edit it, it will show an empty input.

packages/app/components/Today/OpenQueueCard.tsx
```diff
@@ -16,8 +20,23 @@ const AvatarContainer = styled.div`
...
-const OpenQueueCard = ({ queue }: OpenQueueCard) => {
+const EditOutlinedFloatedRight = styled(EditOutlined)`
+  float: right;
+`;
+
+const OpenQueueCard = ({ queue, isTA }: OpenQueueCard) => {
+  const [editingNotes, setEditingNotes] = useState(false);
   const staffList = queue.staffList;
+
+  const updateNotes = (value: any) => {
```
dajinchu, line 31: I think if we are going to take `queue` as a prop we should take a callback `updateNotes` as props as well. The idea being that we should be using swr's `mutate` to make sure to refetch our queue data across the app to avoid any sus inconsistencies. Think about it this way: data should be "owned" by a component. So if the parent component of OpenQueueCard passes us a queue, they should be in charge of updating the queue as well.

see https://github.com/sandboxnu/office-hours/blob/e227ffb7fa8386a25257e466397bd5ba529b243d/packages/app/pages/class/%5Bcid%5D/queue/%5Bqid%5D.tsx#L155 for example

packages/app/components/Today/OpenQueueCard.tsx
willstenzel: Good call. I just changed the input to use `updatedNotes` state instead

packages/app/components/Today/OpenQueueCard.tsx
willstenzel: Gotcha, thanks for the detailed explanation on this one. 

Not sure the best way to mutate the data if the route to get the queue data is `/api/v1/courses/${cid}/queues` but the route to update a queue is `/api/v1/queues/${queueId}`. It looks like use the useSWR `mutate` method needs to be referencing the same key. 

One possibility is to change the queue route on the backend to `/api/v1/queues?course=${cid}` then use the `/api/v1/queues` as the key for the queue data. Would this make sense or is there a better solution?

dajinchu: Yeah, you hit the nail on the head. The example I linked deals with the same issue. We are updating one question of many. That's why we set the new local value of the mutation to:
```
const newQuestions = questions.map((q) =>
  q.id === question.id ? { ...q, status } : q
);
```

NOTE that `OpenQueueCard` has no way to know if we fetched the queue as one of a list or on its own. This is why the component that gave us the queue needs to be in charge of updating it.

dajinchu: Here are the SWR docs: https://swr.vercel.app/docs/mutation#mutate-based-on-current-data
LMK if you want help / pair on this

packages/app/components/Today/OpenQueueCard.tsx
NEUDitao: talked about this earlier with dajinder

packages/app/components/Today/OpenQueueCard.tsx
willstenzel: That's super helpful. I refactored the code to have the `updateQueueNotes` be in the `today.tsx` component and then passed it in as props to the `OpenQueueCard` component. Feels much cleaner now, lmk how it looks

PR #64, opened 2020-07-06
Add settings ui

bugsalexander:
Web notifs not working, but it's because of an issue in master. I will fix in a separate PR.


packages/app/components/Nav/NavBar.tsx
```diff
@@ -8,6 +8,7 @@ import Link from "next/link";
 import { API } from "@template/api-client";
 import useSWR from "swr";
 import { useProfile } from "../../hooks/useProfile";
+import Settings from "./Settings/Profile";
```
aislinblack, line 11: Is there a reason we are importing this? It seems like it isn't being used in this file

packages/app/components/Nav/RightNavBar.tsx
```diff
@@ -1,8 +1,9 @@
 import React, { Component } from "react";
 import styled from "styled-components";
-import { Avatar, Menu } from "antd";
+import { Avatar, Button, Menu } from "antd";
```
aislinblack, line 3: Are we using Button in this?

packages/app/components/Nav/Settings/Profile.tsx
```diff
@@ -0,0 +1,164 @@
...
+              <LableText> Phone </LableText>
+              <InputContainer>
+                {" "}
+                <Input
+                  value={phoneNumber}
+                  onChange={(newNum) => {
+                    setPhoneNumber(newNum.target.value);
+                  }}
+                  placeholder={"XXX-XXX-XXXX"}
+                />{" "}
```
aislinblack, line 141: Do we need the {" "} in this file?

willstenzel:
Does the settings page show on mobile? I looks as if the hamburger menu is covering the avatar icon at mobile widths

image

packages/app/components/Nav/NavBar.tsx
bugsalexander: Good catch! I'm removing all unused imports

packages/app/components/Nav/RightNavBar.tsx
bugsalexander: nope, you're right.

packages/app/components/Nav/Settings/Profile.tsx
bugsalexander: I removed them and the UI looked the same to me, so I don't think so. removing them.

packages/app/components/Nav/Settings/Profile.tsx
```diff
@@ -0,0 +1,164 @@
...
+              <SwitchContainer>
+                {" "}
+                <Switch
+                  checked={phoneNotifToggled}
+                  onChange={onPhoneNotifToggle}
+                />{" "}
+              </SwitchContainer>
+            </Row>
+            <Row>
+              <LableText> Phone </LableText>
```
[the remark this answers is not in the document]
bugsalexander: yeet

packages/app/components/Nav/Settings/Profile.tsx
```diff
@@ -0,0 +1,164 @@
...
+                  value={phoneNumber}
+                  onChange={(newNum) => {
+                    setPhoneNumber(newNum.target.value);
+                  }}
+                  placeholder={"XXX-XXX-XXXX"}
+                />{" "}
+              </InputContainer>
+            </Row>
+            <Row>
+              <Link href="/login" as="/login">
```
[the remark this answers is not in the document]
bugsalexander: thanks for hte tip!

bugsalexander:
@willstenzel I talked to @liustanley about that just now, and I think the icon is actually supposed to be a hamburger like that; however you did catch something else-  we're supposed to be displaying "profile" or "settings" inside the hamburger menu, and we are not.

bugsalexander:
yep @dajinchu . the way olivia and i designed it was such that we display their phone number. of course, it doesn't need to be the case.

bugsalexander:
ready to merge @liustanley

PR #66, opened 2020-07-07
notifRoutes.ts uses the correct ENV

bugsalexander:
this is a hotfix for the .env issue that people were having.

the issue was that while we have a `env.ts` file which exports a global variable to use as the env, not all the files were using it (`notifRoutes.ts`).

additionally, @liustanley found that the TWILIOACCOUNTSID needs to begin with `AC`.

this PR fixes both of the above issues. deploying with `yarn dev` should now work, and not crash, even without a `.env` file in `packages/server` (should use `.env.development` instead).


bugsalexander:
i just received info that most of the stuff in the backend notifyRoutes.ts file will be changing, so I'm closing this PR.

PR #69, opened 2020-07-09
Nest Migration

willstenzel:
It's a chonker of a PR but here it is 😄


packages/server/.env.development
```diff
@@ -1,7 +1,7 @@
 PUBLICKEY=BICM9ljt6UzPC2HsKwlGotdrOSjva59TcwEnolgjf67P8BLN-UStYY2kvTyKDT5DpqJX0sK1z5ZyCdc98_z-AjY
 PRIVATEKEY=3134tddi2nyoxnamvbImuban4fXXgiiXia4t3SGVOBA
 EMAIL=mailto:zuck@fb.com
-TWILIOACCOUNTSID=REMEMBERTHEALAMO
+TWILIOACCOUNTSID=ACREMEMBERTHEALAMO
 TWILIOAUTHTOKEN=FLYMETOTHEMOONANDLETMESINGFOREVERMORE
 TWILIOPHONENUMBER=+14204206969
-COOKIE_PASSWORD=chunguschunguschunguschunguschunguschunguschunguschunguschungus 
\ No newline at end of file
+JWT_SECRET=chunguschunguschunguschunguschunguschunguschunguschunguschungus
```
NEUDitao, line 7: what's the `JWT_SECRET`?

willstenzel: `JWT_SECRET` is the private key used to secure the auth cookie which is set in the user browser

packages/server/README.md
```diff
@@ -0,0 +1,75 @@
+<p align="center">
```
NEUDitao, line 1: can we get rid of this?

packages/server/src/notification/desktop-notif-subscriber.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.notifService = notifService;
+    connection.subscribers.push(this);
+  }
+
+  listenTo() {
+    return DesktopNotifModel;
+  }
+
+  async afterInsert(event: InsertEvent<DesktopNotifModel>) {
+    console.log(`BEFORE DESKTOPNOTIF INSERTED: `, event.entity);
```
liustanley, line 24: Not sure if this console.log is meant for debugging or logging, if it's not logging let's remove it

packages/server/src/queue/queue.entity.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  @Column('text')
+  room: string;
+
+  @OneToMany((type) => QuestionModel, (qm) => qm.queue)
+  @Exclude()
+  questions: QuestionModel[];
+
+  @Expose()
+  get staffList(): UserPartial[] {
```
NEUDitao, line 36: nit: can we just leave a TODO here so we know that this is temp wh ile we figure out OH -> queues?

PR #77, opened 2020-07-14
Main pages UI redesign

liustanley:
Updates the UI styling according to the new Figma mockups: https://www.figma.com/file/tJHzsL30pMnpFAczwJcvji/mockups?node-id=487%3A26334

Only updated `/today`, `/schedule`, and `/queue` for now due to time constraints.


NEUDitao:
can we have screenshots of the new pages?

PR #79, opened 2020-07-15
Cron job

dajinchu:


packages/server/src/course/ical.service.ts
```diff
@@ -42,4 +43,10 @@ export class IcalService {
       officeHours.map((e) => OfficeHourModel.create(e)),
     );
   }
+
+  @Cron(CronExpression.EVERY_DAY_AT_MIDNIGHT)
+  public async updateAllCourses(): Promise<void> {
+    const courses = await CourseModel.find();
+    await Promise.all(courses.map(this.updateCalendarForCourse));
```
willstenzel, line 50: The cron stuff looks nice, just had a question about the `updateCalendarForCoruse` function.

When we call `this.updateCalendarForCoruse` are the old office hours from the course being removed before the new one's are created from the ical?

dajinchu: fixed

PR #81, opened 2020-07-15
use role :)

NEUDitao:
uses role now on queue page


packages/app/pages/course/[cid]/queue/[qid].tsx
```diff
@@ -67,9 +65,6 @@ export default function Queue() {
 
   filterHelpingGroup();
 
-  // Student queue state variables
-  const [isJoining, setIsJoining] = useState<boolean>(false);
```
Comment (NEUDitao) [inline packages/app/pages/course/[cid]/queue/[qid].tsx:71]: wasn't being used

Issue #45, opened 2020-06-29
Associate notifications with a question


NEUDitao:
Yeah... maybe it should be a route on `/queues/:queue_id/questions/:question_id/notify` so that the notification is associated to the question ticket, not a catch all notify-anyone route.

_Originally posted by @dajinchu in https://github.com/sandboxnu/office-hours/pull/33_

PR #33, opened 2020-06-23
Created web push notification features + test buttons on clubs index

NEUDitao:
To set up, you're going to want to generate VAPID keys and put them in an `.env` in `packages/server`. To do so, run `npx web-push generate-vapid-keys` and copy the public and private key into fields in your `.env`. You'll also want to add an email to your `.env`. So in the end, it should look something like:

```
PUBLICKEY=joemama
PRIVATEKEY=sugondesenutsaredelicious
EMAIL=mailto:chu.daj@northeastern.edu
```

If you're using Firefox and are on windows, you're gonna want to make sure that FF is able to interface with Windows. Try the steps [here](https://www.askvg.com/tip-enable-disable-mozilla-firefox-notifications-to-show-in-windows-10-action-center/) if you're not getting notifications.
**Note: the above should not be enabled by default, but for some reason eddys computer is wack**

PR #82, opened 2020-07-15
TAs can notify students

NEUDitao:
Remember the alamo at important.png

### DIFF
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -68,6 +68,9 @@ class APIClient {
       parseQuestionDates(question);
       return question;
     },
+    notify: async (questionId: number): Promise<void> => {
+      await this.axios.post(`/api/v1/questions/${questionId}/notify`);
+    },
   };
   queues = {
     updateNotes: async (queueId: number, notes: string) => {
--- a/packages/app/pages/course/[cid]/queue/[qid].tsx
+++ b/packages/app/pages/course/[cid]/queue/[qid].tsx
@@ -137,9 +137,8 @@ export default function Queue() {
     mutate(`/api/v1/queues/${qid}/questions`, newQuestions);
   };
 
-  // TODO: yup
-  const alertStudent = (question: Question) => {
-    // Send API request to trigger notification
+  const alertStudent = async (question: Question) => {
+    await API.questions.notify(question.id);
   };
 
   if (questions) {
--- a/packages/server/src/notification/notification.controller.ts
+++ b/packages/server/src/notification/notification.controller.ts
@@ -1,7 +1,7 @@
-import { Controller, Get, Body, Post, Param, UseGuards } from '@nestjs/common';
+import { Body, Controller, Get, Param, Post, UseGuards } from '@nestjs/common';
 import { DesktopNotifBody } from '@template/common';
-import { NotificationService } from './notification.service';
 import { JwtAuthGuard } from '../profile/jwt-auth.guard';
+import { NotificationService } from './notification.service';
 
 @Controller('notifications')
 @UseGuards(JwtAuthGuard)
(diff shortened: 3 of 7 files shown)


packages/server/src/notification/notification.controller.ts
```diff
@@ -37,9 +37,13 @@ export class NotificationController {
     return `registration success for ${body.phoneNumber}`;
   }
 
+  // TODO: GET RID OF THIS
   @Post('/notify_user/:user_id')
   async notifyUser(@Param('user_id') user_id: number): Promise<string> {
-    await this.notifService.notifyUser(user_id);
+    await this.notifService.notifyUser(
+      user_id,
+      "bruh if you're still tryna test notifications, ligma balls, FUCK FUCK FUCK FUCK FUCK, sugondese, hot cheetos",
```
NEUDitao, line 45: this is the test endpoint. we decided to put in all of the "hot words" that we're bound to search for when we do a final cleanup

packages/server/src/question/question.controller.ts
```diff
@@ -150,4 +154,34 @@ export class QuestionController {
...
+      })) === 1;
+
+    if (!isUserTAOfCourse) {
+      throw new UnauthorizedException(
+        "bruh you ain't a TA no spamming peeps for u xD",
+      );
+    }
+
+    this.notifService.notifyUser(
+      question.creatorId, // TODO: Stanley think of a better message
```
NEUDitao, line 183: @liustanley

packages/api-client/index.ts
```diff
@@ -68,6 +68,9 @@ class APIClient {
       parseQuestionDates(question);
       return question;
     },
+    notify: async (questionId: number): Promise<void> => {
```
liustanley, line 71: Can this just be `: void` instead? Not sure if that messes something up

packages/server/src/notification/notification.service.ts
```diff
@@ -50,7 +50,7 @@ export class NotificationService {
   }
 
   // Notify user on all platforms
-  async notifyUser(userId: number) {
+  async notifyUser(userId: number, message: string): Promise<void> {
```
liustanley, line 53: Same thing here? Again, not sure if making the return type void will break the async

packages/server/src/question/question.controller.ts
```diff
@@ -150,4 +154,34 @@ export class QuestionController {
...
+        where: {
+          role: Role.TA,
+          courseId: question.queue.courseId,
+          userId: userId,
+        },
+      })) === 1;
+
+    if (!isUserTAOfCourse) {
+      throw new UnauthorizedException(
+        "bruh you ain't a TA no spamming peeps for u xD",
```
liustanley, line 178: xd

packages/app/pages/course/[cid]/queue/[qid].tsx
```diff
@@ -137,9 +137,8 @@ export default function Queue() {
     mutate(`/api/v1/queues/${qid}/questions`, newQuestions);
   };
 
-  // TODO: yup
-  const alertStudent = (question: Question) => {
-    // Send API request to trigger notification
+  const alertStudent = async (question: Question) => {
+    await API.questions.notify(question.id);
```
Comment (liustanley) [inline packages/app/pages/course/[cid]/queue/[qid].tsx:141]: is there a reason why we can't use the notification controller for this alert? we have the student's id from the question's creator property, we would just have to add a request body to store the message. i think this would also remove the need for a new endpoint in question controller

Comment (liustanley) [reply packages/app/pages/course/[cid]/queue/[qid].tsx:141]: nvm! just read the discussion on #33 and the issue #45 

makes sense to have a question notification endpoint to validate that the TA trying to alert is currently helping/in the queue

packages/server/src/question/question.controller.ts
```diff
@@ -150,4 +154,34 @@ export class QuestionController {
...
+
+  @Post(':questionId/notify')
+  async notify(
+    @Param('questionId') questionId: number,
+    @UserId() userId: number,
+  ): Promise<void> {
+    const question = await QuestionModel.findOne(questionId, {
+      relations: ['queue'],
+    });
+
```
liustanley, line 166: can we add a TODO here to check that the user is a TA of the question's queue or currently helping that question?

packages/api-client/index.ts
dajinchu: it gotta be a promise

PR #83, opened 2020-07-15
isTA is GONE

NEUDitao:
yup

### DIFF
--- a/packages/app/pages/course/[cid]/queue/[qid].tsx
+++ b/packages/app/pages/course/[cid]/queue/[qid].tsx
@@ -16,6 +16,7 @@ import StudentPopupCard from "../../../../components/Queue/StudentPopupCard";
 import StudentQueueList from "../../../../components/Queue/StudentQueueList";
 import TAQueueList from "../../../../components/Queue/TAQueueList";
 import { useProfile } from "../../../../hooks/useProfile";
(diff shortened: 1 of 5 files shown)


packages/app/pages/course/[cid]/queue/[qid].tsx
```diff
@@ -28,7 +29,7 @@ export default function Queue() {
   const profile = useProfile();
   const router = useRouter();
   const { cid, qid } = router.query;
-  const role = profile?.courses.find((e) => e.course.id === Number(cid)).role;
+  const role = useRoleInCourse(Number(cid));
```
Comment (willstenzel) [inline packages/app/pages/course/[cid]/queue/[qid].tsx:32]: Do we want to have `useRoleInCourse` hook do the casting of the `cid` from a string to a number so that we don't have to remember to cast it each time we call the function?

packages/app/pages/course/[cid]/today.tsx
```diff
@@ -68,7 +67,7 @@ export default function Today() {
             <Col md={12} xs={24}>
               <Row justify="space-between">
                 <Title>Current Office Hours</Title>
-                {isTA && (
+                {role === Role.TA && (
```
Comment (liustanley) [inline packages/app/pages/course/[cid]/today.tsx:70]: fuck i should've looked at this before committing to #81 😭

Comment (NEUDitao) [reply packages/app/pages/course/[cid]/today.tsx:70]: your mom should've looked at this before committing to #81

PR #87, opened 2020-07-17
Ta check in

NEUDitao:
image

image
image

and then redirects you

### DIFF
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -5,6 +5,7 @@ import {
   GetCourseResponse,
   GetProfileResponse,
   GetQuestionResponse,
+  GetQueueResponse,
   ListQuestionsResponse,
   PhoneNotifBody,
   TAUpdateStatusResponse,
@@ -80,6 +81,9 @@ class APIClient {
     },
   };
   queues = {
+    get: async (queueId: number): Promise<GetQueueResponse> => {
+      return await this.axios.get(`/api/v1/queues/${queueId}`);
+    },
     updateNotes: async (queueId: number, notes: string) => {
       await this.axios.patch(`/api/v1/queues/${queueId}`, { notes });
     },
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -30,13 +30,20 @@ class APIClient {
     },
   };
   taStatus = {
-    update: async (courseId: number): Promise<TAUpdateStatusResponse> => {
+    checkIn: async (
+      courseId: number,
+      room: string
+    ): Promise<TAUpdateStatusResponse> => {
       const queue = (
-        await this.axios.patch(`/api/v1/courses/${courseId}/ta/change_status`)
+        await this.axios.post(`/api/v1/courses/${courseId}/ta_location/${room}`)
       ).data;
-      parseQueueDates(queue);
-      queue.questions.forEach((question: any) => parseQuestionDates(question));
       return queue;
+    },
+    delete: async (courseId: number, room: string): Promise<void> => {
+      //TODO:
+      await this.axios.delete(
+        `/api/v1/courses/${courseId}/ta_location/${room}`
+      );
     },
   };
   questions = {
--- a/packages/app/components/Today/TACheckinButton.tsx
+++ b/packages/app/components/Today/TACheckinButton.tsx
@@ -22,6 +22,9 @@ export default function TACheckinButton({
 
   const [viewCheckinModal, setViewCheckinModal] = useState(false);
   const [value, setValue] = useState(0);
+  const [customRoom, setCustomRoom] = useState("");
+
+  const canSubmitCustomRoom = !(value === -1 && !customRoom);
 
   const { data } = useSWR(
     `api/v1/courses/${courseId}`,
@@ -33,6 +36,20 @@ export default function TACheckinButton({
     height: "30px",
     lineHeight: "30px",
   };
+
+  async function checkInTA() {
+    if (canSubmitCustomRoom) {
+      const redirectID = await API.taStatus.checkIn(
+        courseId,
+        value === -1 ? customRoom : data?.queues[value].room
+      );
+
+      router.push(
+        "/course/[cid]/queue/[qid]",
+        `/course/${courseId}/queue/${redirectID.id}`
+      );
+    }
+  }
 
   return (
     <>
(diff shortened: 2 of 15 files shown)


packages/app/pages/course/[cid]/queue/[qid].tsx
```diff
@@ -32,16 +32,17 @@ export default function Queue() {
...
   );
-  const { data: course, error: queuesError } = useSWR(
-    qid ? `/api/v1/courses/${cid}` : null,
-    async () => API.course.get(Number(cid))
+
+  const { data: queue, error: queuesError } = useSWR(
+    qid && `/api/v1/queues/${qid}`,
+    async () => API.queues.get(Number(qid))
   );
+  // TODO: put room name on this page somewhere so we know where we are :)
```
Comment (NEUDitao) [inline packages/app/pages/course/[cid]/queue/[qid].tsx:43]: this gets removed in the next PR cause Dajin and I have smooth brains

packages/api-client/index.ts
```diff
@@ -30,14 +31,21 @@ class APIClient {
...
   };
   taStatus = {
-    update: async (courseId: number): Promise<TAUpdateStatusResponse> => {
+    checkIn: async (
+      courseId: number,
+      room: string
+    ): Promise<TAUpdateStatusResponse> => {
       const queue = (
-        await this.axios.patch(`/api/v1/courses/${courseId}/ta/change_status`)
+        await this.axios.post(`/api/v1/courses/${courseId}/ta_location/${room}`)
```
willstenzel, line 39: It might make sense here to put the room in the body of the request here. When I tested it out I got a request url like `/api/v1/courses/1/ta_location/RY%20154`

packages/api-client/index.ts
```diff
@@ -30,14 +31,21 @@ class APIClient {
...
       ).data;
-      parseQueueDates(queue);
-      queue.questions.forEach((question: any) => parseQuestionDates(question));
       return queue;
     },
+    delete: async (courseId: number, room: string): Promise<void> => {
+      //TODO:
+      await this.axios.delete(
+        `/api/v1/courses/${courseId}/ta_location/${room}`
```
willstenzel, line 46: Same as above

packages/app/components/Today/TACheckinButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const [viewCheckinModal, setViewCheckinModal] = useState(false);
+  const [value, setValue] = useState(0);
+  const [customRoom, setCustomRoom] = useState("");
+
+  const canSubmitCustomRoom = !(value === -1 && !customRoom);
+
+  const { data } = useSWR(
+    `api/v1/courses/${courseId}`,
+    async () => courseId && API.course.get(Number(courseId))
```
willstenzel, line 31: Minor thing but if the props types expect this to be a number do we need to cast it in `API.course.get(Number(courseId))`?

packages/app/components/Today/TACheckinButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <CheckinButton
+        type="default"
+        size="large"
+        onClick={() => setViewCheckinModal(true)}
+        disabled={!data}
+      >
+        Check In
+      </CheckinButton>
+      <Modal
+        title="Time to Check In! :D 🐱o(=•ェ•=)m"
```
willstenzel, line 65: Don't know what this emoji means but I like it 👍

packages/server/src/course/course.controller.ts
```diff
@@ -20,7 +26,51 @@ export class CourseController {
...
+    );
+
+    if (!queue) {
+      queue = await QueueModel.create({
+        room,
+        courseId,
+        staffList: [],
+      }).save();
+    }
+
```
willstenzel, line 70: Do we need to check that the user is not already in the `staffList` of the queue? Just incase they hit "Check in" twice?

packages/server/src/profile/user.entity.ts
```diff
@@ -40,4 +42,7 @@ export class UserModel extends BaseEntity {
   @OneToMany((type) => PhoneNotifModel, (notif) => notif.user)
   @Exclude()
   phoneNotifs: PhoneNotifModel[];
+
```
willstenzel, line 45: Should we add Exclude here if the value is not being used on the frontend?
```suggestion

@Exclude()
```

packages/api-client/index.ts
NEUDitao: dajin and I were discussing it and I dont' remember why we decided to leave it in the URL but there was a reason. I think it was just cause Nest auto encode/decodes the URLs sent and doing it in a body at that point seemed to be more squeeze for the same quality of juice

packages/app/components/Today/TACheckinButton.tsx
NEUDitao: :).

packages/app/components/Today/TACheckinButton.tsx
NEUDitao: if you're on windows, and type in chinese, typing animals gives these weird emojis

packages/server/src/course/course.controller.ts
NEUDitao: typeorm is smart enough for this :o

packages/server/src/profile/user.entity.ts
NEUDitao: probably

packages/api-client/index.ts
dajinchu: url encoding is not a bad thing typically. It's not super important where it goes, however, `DELETE` http requests aren't supposed to put stuff in the request body, so it'd be better to put it in the path for when we check out via `DELETE`

packages/app/components/Today/TACheckinButton.tsx
```diff
@@ -0,0 +1,94 @@
+import { API } from "@template/api-client";
+import { Button, Input, Modal, Radio } from "antd";
+import { useRouter } from "next/router";
+import { ReactElement, useState } from "react";
+import styled from "styled-components";
+import useSWR from "swr";
+
+const CheckinButton = styled(Button)`
```
liustanley, line 8: nice CSS 😎

packages/app/pages/course/[cid]/today.tsx
```diff
@@ -68,11 +60,7 @@ export default function Today() {
...
                 <Title>Current Office Hours</Title>
-                {role === Role.TA && (
-                  <CreateQueueButton type="default" size="large">
-                    Create Queue
-                  </CreateQueueButton>
-                )}
+                {role === Role.TA && <TACheckinButton courseId={Number(cid)} />}
               </Row>
               {data?.queues?.map((q) => (
                 <OpenQueueCard
```
Comment (liustanley) [inline packages/app/pages/course/[cid]/today.tsx:66]: jk i did have one thing, it seems when i join a queue as a student, the # in queue counter on the today page doesn't update now, i believe this was working before tho

packages/app/pages/course/[cid]/queue/[qid].tsx
```diff
@@ -32,16 +32,17 @@ export default function Queue() {
   const role = useRoleInCourse(Number(cid));
```
Comment (liustanley) [inline packages/app/pages/course/[cid]/queue/[qid].tsx:33]: oh one more thing! can we prevent TAs from helping students if they're not checked in? we could just disable the student sidebar buttons for when they're not checked in

PR #88, opened 2020-07-17
Ta check out

NEUDitao:
image
image

### DIFF
--- a/packages/server/src/course/course.controller.ts
+++ b/packages/server/src/course/course.controller.ts
@@ -1,6 +1,7 @@
 import {
   ClassSerializerInterceptor,
   Controller,
+  Delete,
   Get,
   Param,
   Post,
@@ -72,5 +73,23 @@ export class CourseController {
     await queue.save();
 
     return queue;
+  }
+
+  @Delete(':id/ta_location/:room')
+  async checkOut(
+    @Param('id') courseId: number,
+    @Param('room') room: string,
+    @User() user: UserModel,
+  ): Promise<void> {
+    const queue = await QueueModel.findOne(
+      {
+        room,
+        courseId,
+      },
+      { relations: ['staffList'] },
+    );
+
+    queue.staffList = queue.staffList.filter((e) => e.id !== user.id);
+    await queue.save();
   }
 }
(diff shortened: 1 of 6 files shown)


NEUDitao:
PLEASE change the base to master when we're merging ty

packages/server/test/course.integration.ts
```diff
@@ -71,4 +72,52 @@ describe('Course Integration', () => {
...
+      const ta = await UserFactory.create();
+      const queue = await QueueFactory.create({ room: 'The Alamo' });
+      const tcf = await TACourseFactory.create({
+        course: queue.course,
+        user: ta,
+      });
+
+      await supertest({ userId: ta.id })
+        .delete(`/courses/${tcf.courseId}/ta_location/The Alamo`)
+        .expect(200);
```
willstenzel, line 114: Title implies the response should return an error but the expected response code is a 200. Should the response be something like `404 Bad Request` instead?

packages/server/src/course/course.controller.ts
```diff
@@ -73,4 +74,22 @@ export class CourseController {
...
+    @User() user: UserModel,
+  ): Promise<void> {
+    const queue = await QueueModel.findOne(
+      {
+        room,
+        courseId,
+      },
+      { relations: ['staffList'] },
+    );
+
```
willstenzel, line 91: If the ta is not in the current `staffList` for the queue should we return a response so the front end knows the user wasn't actually checked out?

packages/server/src/course/course.controller.ts
```diff
@@ -73,4 +74,22 @@ export class CourseController {
 
     return queue;
   }
+
+  @Delete(':id/ta_location/:room')
```
willstenzel, line 78: I mentioned this in the other PR but it might better to have the room in the payload rather than the url. It works either way though, I'm curious what other people think

packages/server/test/course.integration.ts
NEUDitao: OOPS

packages/server/src/course/course.controller.ts
NEUDitao: I think it's fine cause either way the TA isn't in the queue

packages/server/test/course.integration.ts
dajinchu: i think it might be better with a `4xx` but it doesn't actually impact us either way

Issue #76, opened 2020-07-13
Allow TAs to open scheduled office hours 15 minutes in advance


liustanley:
- Create alternate OpenQueueBlock on the Today Page that allows TAs to open a scheduled queue 15 minutes early.
- Change backend OfficeHours query to include queues 15 minutes ahead, maybe mark each office hours object with a field to indicate that it's early?

PR #89, opened 2020-07-18
Linked Office Hours to queues using rooms and time

NEUDitao:
remember to make this point to master when you merge


packages/server/src/course/course.controller.ts
```diff
@@ -26,9 +27,46 @@ export class CourseController {
...
+      },
+      relations: ['staffList'],
+    });
+
+    const nonEmptyQueues = queues.filter((e) => e.staffList.length > 0);
+
+    const queuesHappeningNow = [];
+
+    for (const q of queues) {
+      const oh = course.officeHours.find((f) => f.room === q.room);
```
willstenzel, line 58: Should this line reference the `rooms` const from before?
```suggestion
      const oh = rooms.find((f) => f.room === q.room);
```

packages/server/src/course/course.controller.ts
```diff
@@ -26,9 +27,46 @@ export class CourseController {
...
+      .map((e) => e.room);
+
+    const queues = await QueueModel.find({
+      where: {
+        courseId: id,
+      },
+      relations: ['staffList'],
+    });
+
+    const nonEmptyQueues = queues.filter((e) => e.staffList.length > 0);
```
willstenzel, line 53: I think this is the best way of doing things for now but I'm still not sure how we'll be able to return the upcoming office hours from #76 if the upcoming hours don't have any staff in them yet?
Maybe we'll have another endpoint to return the upcoming office hours

packages/server/src/queue/queue.entity.ts
```diff
@@ -42,6 +42,11 @@ export class QueueModel extends BaseEntity {
   @JoinTable()
   staffList: UserModel[];
 
+  time?: {
```
willstenzel, line 45: This is slick, nice work

packages/server/src/course/course.controller.ts
willstenzel, line 58: Instead of doing the `find` across all officeHours don't we want to be looking through only the officeHours that are currently happening? (like you do on line 39 with the rooms `filter`)

willstenzel:
Sorry I accidentally deleted a review comment so that’s why had to request changes twice

packages/server/src/course/course.controller.ts
NEUDitao: the upcoming queues are returned cause of the funky math tho

packages/server/src/course/course.controller.ts
NEUDitao: whatchu think @dajinchu

packages/server/src/course/course.controller.ts
```diff
@@ -26,9 +27,44 @@ export class CourseController {
...
+
+    const queuesHappeningNow = [];
+
+    for (const q of queues) {
```
willstenzel, line 55: Do we want to be looping through the list of `officeHoursHappeningNow` here so that if a queue does not exist for one of the office hours that currently happening, we can create new queue?

NEUDitao: mans got bigger brains than us @dajinchu

dajinchu: I thought we should do this when we scrape ical

willstenzel: That totally works. In that case we can probably just iterate through the `officeHoursHappeningNow` instead of all the queues in a course, since we'll know each office hours has a particular `queue` with the same room associated with it. 

Also if we go that route we could also remove the `if (oh) {` part.

packages/app/components/Today/OpenQueueCard.tsx
```diff
@@ -99,7 +101,19 @@ const OpenQueueCard = ({ queue, isTA, updateQueueNotes }: OpenQueueCard) => {
     <PaddedCard
       headStyle={{ background: "#F3F5F7" }}
       title={staffList.map((staffMember) => staffMember.name).join(", ")}
-      extra={<ExtraText>//TODO: 3:00 - 5:00</ExtraText>}
+      extra={
```
liustanley, line 104: This can be done later, but since we're converting to local time, it might cause confusion for students who assume all the times are EST. Maybe we could add a hover popup that explains these times are displayed in your local time?

packages/server/src/course/course.controller.ts
NEUDitao: cool pls approve

packages/app/components/Today/OpenQueueCard.tsx
NEUDitao: added a TODO

PR #91, opened 2020-07-18
Twilio verification

NEUDitao:
When you sign up for phone numbers it now asks you if you want to sign up or not. It's kind of tricky to test but lmk if you want to see a demo/try it out for yourself.

### DIFF
--- a/packages/server/src/notification/notification.controller.ts
+++ b/packages/server/src/notification/notification.controller.ts
@@ -1,19 +1,31 @@
-import { Body, Controller, Get, Param, Post, UseGuards } from '@nestjs/common';
+import {
+  Body,
+  Controller,
+  Get,
+  Param,
+  Post,
+  Req,
+  Res,
+  UseGuards,
+} from '@nestjs/common';
 import { DesktopNotifBody } from '@template/common';
+import { Request, Response } from 'express';
 import { JwtAuthGuard } from '../profile/jwt-auth.guard';
 import { NotificationService } from './notification.service';
+import { PhoneNotifModel } from './phone-notif.entity';
 
 @Controller('notifications')
-@UseGuards(JwtAuthGuard)
 export class NotificationController {
   constructor(private notifService: NotificationService) {}
 
   @Get('desktop/credentials')
+  @UseGuards(JwtAuthGuard)
   getDesktopCredentials(): string {
     return JSON.stringify(this.notifService.desktopPublicKey);
   }
 
   @Post('desktop/register/:user_id')
+  @UseGuards(JwtAuthGuard)
   async registerDesktopUser(
     @Body() body: DesktopNotifBody,
     @Param('user_id') user_id: number,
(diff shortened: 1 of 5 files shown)


packages/server/src/notification/notification.controller.ts
```diff
@@ -1,19 +1,30 @@
...
 import { JwtAuthGuard } from '../profile/jwt-auth.guard';
 import { NotificationService } from './notification.service';
 
 @Controller('notifications')
-@UseGuards(JwtAuthGuard)
 export class NotificationController {
   constructor(private notifService: NotificationService) {}
 
   @Get('desktop/credentials')
+  @UseGuards(JwtAuthGuard)
```
NEUDitao, line 21: this is so there's no guard on the verify endpoint , so twilio can actually interact with our API

packages/server/src/notification/notification.service.ts
```diff
@@ -88,16 +95,70 @@ export class NotificationService {
...
-          body: message,
-          from: this.configService.get('TWILIOPHONENUMBER'),
-          to: pn.phoneNumber,
-        }));
-    } catch (error) {
-      console.error('problem sending message', error);
+  async notifyPhone(
+    pn: PhoneNotifModel,
+    message: string,
+    verificationMethod: boolean,
```
dajinchu, line 101: Confusing variable. Maybe call it `force`

packages/server/src/notification/notification.service.ts
```diff
@@ -88,16 +95,70 @@ export class NotificationService {
...
+    phoneNumber: string,
+    message: string,
+    response: Response,
+  ): Promise<void> {
+    const MessagingResponse = require('twilio').twiml.MessagingResponse;
+    const twiml = new MessagingResponse();
+
+    if (message !== 'YES' && message !== 'NO' && message !== 'STOP') {
+      twiml.message(
+        'Please respond with either YES or NO. Text STOP at any time to stop receiving text messages',
```
dajinchu, line 127: Can we put all these messages in a constants file so that we can edit the copy easier

packages/server/src/notification/notification.service.ts
```diff
@@ -88,16 +95,70 @@ export class NotificationService {
...
+      } catch (error) {
+        console.error('problem sending message', error);
+      }
+    }
+  }
+
+  async verifyPhone(
+    phoneNumber: string,
+    message: string,
+    response: Response,
```
dajinchu, line 120: Services should not deal with the http layer. I think you could have this function return a string and have the controller send that string back and set the content type

packages/server/src/notification/notification.controller.ts
```diff
@@ -39,11 +51,23 @@ export class NotificationController {
...
   async notifyUser(@Param('user_id') user_id: number): Promise<string> {
     await this.notifService.notifyUser(
       user_id,
       "bruh if you're still tryna test notifications, ligma balls, FUCK FUCK FUCK FUCK FUCK, sugondese, hot cheetos",
     );
     return 'notified';
   }
+
+  @Post('/phone/verify')
+  async verifyPhoneUser(
```
dajinchu, line 64: Can we use `@Body` instead of getting the request?

Also note that we can set the content type header using a decorator https://docs.nestjs.com/controllers#headers

Then I don't think we need the req and res at all

dajinchu, line 64: Also I think we could add a comment making it more obvious this is a webhook endpoint that comes from twilio and linking their docs. Also I'm wondering if twilio has a way for us to verify their identity? What's stopping me from postmanning this endpoint with Alex's phone number?

NEUDitao: we need res to send a message back right? i'm not too familiar with express/nest conventions and crossing still

dajinchu: Nest lets us just return the data we wanna send.
For example:
https://github.com/sandboxnu/office-hours/blob/f3147bb0fb51674cb5565277255ab329d40fbe8b/packages/server/src/course/course.controller.ts#L21-L22

PR #97, opened 2020-07-20
Pm2

dajinchu:
Deploy to staging and production with pm2. GitHub action auto-deploys master to staging.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,23 @@
+name: Deploy Staging
+
+on:
+  push:
+    branches:
+      - pm2
```
liustanley, line 6: should this be changed now that we're merging pm2?

packages/server/package.json
```diff
@@ -7,11 +7,10 @@
   "license": "UNLICENSED",
   "scripts": {
     "prebuild": "rimraf dist",
-    "build": "nest build",
-    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
-    "dev": "nest start --watch --preserveWatchOutput",
+    "build": "nest build --webpack --webpackPath webpack.config.js",
+    "dev": "nest build --webpack --webpackPath webpack-hmr.config.js",
```
liustanley, line 11: should this be `nest start` instead of build since we're merging into master?

dajinchu: no, it's weird but the `dev` script works because webpack-hmr.config.js  makes it so the build process kicks off the server. However, without the hmr config, it just builds.

.github/workflows/deploy.yml
dajinchu: good catch

packages/server/package.json
liustanley: interesting ok! that's fine then

PR #100, opened 2020-07-20
TA's that aren't checked in can't help students

NEUDitao:
TODO: fix the crap outta student cards

image
image


packages/app/components/Queue/QueueCardSharedComponents.tsx
```diff
@@ -0,0 +1,51 @@
+import { Question } from "@template/common";
```
NEUDitao, line 1: the queuecards are all refactors to make finding code way easier

packages/app/components/Queue/TAQueueList.tsx
```diff
@@ -270,8 +298,20 @@ export default function TAQueueList({
...
-              {queue?.staffList.some((e) => e.id === user.id) ? (
+              <Tooltip
+                title={
+                  !isStaffCheckedIn && "You must check in to help students!"
+                }
+              >
+                <HelpNextButton
+                  disabled={!isStaffCheckedIn}
+                  size="large"
+                  tooltip="test"
```
willstenzel, line 309: Is this needed?

NEUDitao: prob not... will get to after work and then merge

PR #101, opened 2020-07-20
Ta delete confirmation

NEUDitao:
image

image

### DIFF
--- a/packages/app/components/Queue/UpdateQuestionPopover.tsx
+++ b/packages/app/components/Queue/UpdateQuestionPopover.tsx
@@ -1,0 +1,23 @@
+import { Popconfirm } from "antd";
+import { ReactElement } from "react";
+
+interface UpdateQuestionPopoverProps {
+  onConfirm: () => void;
+  children: ReactElement;
+}
+
+export default function UpdateQuestionPopover({
+  onConfirm,
+  children,
+}: UpdateQuestionPopoverProps): ReactElement {
+  return (
+    <Popconfirm
+      title="Are you sure you want to delete this question from the queue?"
+      okText="Yes"
+      cancelText="No"
+      onConfirm={onConfirm}
+    >
+      {children}
+    </Popconfirm>
+  );
+}
--- a/packages/app/components/Queue/StudentInfoCard.tsx
+++ b/packages/app/components/Queue/StudentInfoCard.tsx
@@ -1,12 +1,13 @@
-import React from "react";
-import styled from "styled-components";
-import { Card, Row, Col, Avatar, Button } from "antd";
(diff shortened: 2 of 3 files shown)


packages/app/components/Queue/UpdateQuestionPopover.tsx
```diff
@@ -0,0 +1,23 @@
+import { Popconfirm } from "antd";
+import { ReactElement } from "react";
+
+interface UpdateQuestionPopoverProps {
+  onConfirm: () => void;
+  children: ReactElement;
+}
+
+export default function UpdateQuestionPopover({
```
dajinchu, line 9: Not 100% sure this component is worth it. It's not really doing anything?

packages/app/components/Queue/UpdateQuestionPopover.tsx
```diff
@@ -0,0 +1,23 @@
...
+  onConfirm,
+  children,
+}: UpdateQuestionPopoverProps): ReactElement {
+  return (
+    <Popconfirm
+      title="Are you sure you want to delete this question from the queue?"
```
willstenzel, line 15: Should this title be passed in as `props` so it can be set dynamically for the use case?

When a TA clicks the can't find button, it could be slightly confusing to see the word "delete" in the title of the Popover. Maybe we could say something like 'Are you sure you want to mark this question as "Can't Find"?'

packages/app/components/Queue/UpdateQuestionPopover.tsx
NEUDitao: matthias came into my brian and told me "you have common functionality," but I'm aight with removing it if anyone else has a strong opinion

aislinblack: "into my brian"

NEUDitao: shhh i'm TRYING

packages/app/components/Queue/UpdateQuestionPopover.tsx
NEUDitao: sure... might just remove `UpdateQuestionPopover` and then just use normal Popconfirm at that point cause that's a little too much dryness or w/e

liustanley:
Not sure if you already added this, but we also want a confirmation on Students choosing to leave/delete their question

packages/app/components/Queue/UpdateQuestionPopover.tsx
willstenzel: Cool, that sounds good

PR #102, opened 2020-07-20
remove checked in staff if no staff are checked in

NEUDitao:
image


dajinchu:
We might wanna come up with clearer text. It's not obvious to a student whether or not they can/should join a queue that has no staff checked in. Feel free to merge this tho, maybe just add a TODO. We can fix all the text later

NEUDitao:
> We might wanna come up with clearer text. It's not obvious to a student whether or not they can/should join a queue that has no staff checked in. Feel free to merge this tho, maybe just add a TODO. We can fix all the text later

i'll add the todo after work and then merge

PR #104, opened 2020-07-20
Question draft in progress indicator

aislinblack:
Screen Shot 2020-07-20 at 11 02 35 AM

Testing steps:
- Go to the queue Page
- You should not have a draft in local storage, so do not expect the popup. 
- Start to edit your question, do not finish question, reload page
- Note the popup, click continue drafting question. See your question from before. Then do not finish again and reload the page
- Delete question, refresh. (no popup)


packages/app/components/Queue/QuestionForm.tsx
```diff
@@ -66,13 +67,38 @@ export default function QuestionForm({
   // on question type change, update the question type state
   const onCategoryChange = (e: RadioChangeEvent) => {
     setQuestionTypeInput(e.target.value);
+
+    const questionFromStorage = JSON.parse(
+      window.localStorage.getItem("draftQuestion")
```
NEUDitao, line 72: why do you use `localStorage` instead of state?

dajinchu: wait a second. I think it'd make most sense to get it from the server. Currently, when you draft a question, we create a question server side and set its status to `Drafting` so we should be able to just figure out whether you have a draft in progress by looking through the question list and filtering for questions created by current user with status = drafting.

dajinchu: No local state should be used at all, because we're just synching with the server.

aislinblack: But I don't think it makes any sense to save a question to the serve that many times. Like sure for the category change it might make sense. But when they're midtyping their question and then the internet cuts out or their browser just shuts down? @liustanley, @ofloody, and I discussed this before starting that it makes sense for the server to know that there is a draft question, but for us not to back and forth save the question repeatedly.

NEUDitao: i agree with aislin, pinging the server on every value change to the input box seems excessive

liustanley:
> Don't think this was introduced by your PR but I was curious from a UX standpoint do we want to show "Your Question" question section if the student is still drafting the question?
> 
> @liustanley, or anyone else, feel free to weigh in on this
> 
> image

We decided to hide the "Your Question" card if you're currently drafting a question.

aislinblack:
> > Don't think this was introduced by your PR but I was curious from a UX standpoint do we want to show "Your Question" question section if the student is still drafting the question?
> > @liustanley, or anyone else, feel free to weigh in on this
> > image
> 
> We decided to hide the "Your Question" card if you're currently drafting a question.

Just put in a quick fix that should resolve this

dajinchu:
#121 adds a hook so y'all can rebase off it

packages/app/components/Queue/QuestionForm.tsx
```diff
@@ -49,7 +50,12 @@ export default function QuestionForm({
   question,
   leaveQueue,
   finishQuestion,
-}: QuestionFormProps) {
+}: QuestionFormProps): JSX.Element {
```
NEUDitao, line 53: I think `ReactElement` is the one we've been using across the whole codebase

packages/app/components/Queue/QuestionForm.tsx
```diff
@@ -66,13 +72,35 @@ export default function QuestionForm({
   // on question type change, update the question type state
   const onCategoryChange = (e: RadioChangeEvent) => {
     setQuestionTypeInput(e.target.value);
+
+    const questionFromStorage = storageQuestion ?? {};
+
+    window.localStorage.setItem(
```
NEUDitao, line 78: can't you just `setStoredQuestion` here? I dont' know all of the details of the hook tho tbh

packages/app/hooks/useLocalStorage.ts
```diff
@@ -26,5 +26,15 @@ export function useLocalStorage<T>(
...
   };
 
-  return [storedValue, setValue];
+  const removeValue = () => {
+    try {
+      setStoredValue(null);
+
+      window.localStorage.removeItem(key);
+    } catch (error) {
+      console.log(error);
```
NEUDitao, line 35: nit, can we use `console.error` for logging errors?

PR #116, opened 2020-07-20
Cleanup SWR

dajinchu:
## Before 
swr_today_old 

## After
swr_today_stable 

These are little details but basically: 
- Layout doesn't shift down on load
- Load calendar and card shell before data comes in.


packages/app/pages/course/[cid]/queue/[qid].tsx
```diff
@@ -31,7 +31,7 @@ export default function Queue() {
   const { cid, qid } = router.query;
   const role = useRoleInCourse(Number(cid));
 
-  const { data: questions, error: questionsError } = useSWR(
+  const { data: questions = [], error: questionsError } = useSWR(
```
Comment (dajinchu) [inline packages/app/pages/course/[cid]/queue/[qid].tsx:34]: this lets us render subcomponent's shells before data arrives

PR #117, opened 2020-07-20
Update when taHelped is set and add extra checks

### LINKED ISSUE #115: TA question ownership

willstenzel:
Previously the taHelped was set when the Question status was changed to `Resolved` however, this does not allow for TA's to have "ownership" over particular students they are helping at any given time. 

With this PR we'll now be able to know when TA's are currently helping a student, so in the case the TA looses connection along with their local state on the frontend, we'll still be able to know which TA's are helping which students. 

Closes #115


liustanley:
> I'm good with this -- but should we consider storing who the TA is helping in the DB?

We're storing who the TA is helping through `question.taHelped`

PR #118, opened 2020-07-21
Fix TSC

dajinchu:
- Get TSC script to work and not blow up on the decorators error
- Make eslint only lint `.ts` and `.tsx`
- GitHub action prettier check everything in the branch

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,11 +1,18 @@
-name: Lint
+name: Lint and Typescript
 
 on: pull_request
 
 jobs:
-  lint:
-    name: lint
+  setup:
+    name: setup
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        cmd:
+          - "yarn pretty-quick --branch --check"
+          - "yarn lint"
+          - "yarn tsc"
     steps:
       - uses: actions/checkout@v2
       - name: install node v12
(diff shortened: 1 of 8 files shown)


.github/workflows/lint.yml
```diff
@@ -1,17 +1,23 @@
...
 on: pull_request
 
 jobs:
-  lint:
-    name: lint
+  setup:
+    name: setup
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
dajinchu, line 10: This is interesting, it makes it so all three commands are run, even if one of them fails. That way we always get the full lint/prettier/eslint output

package.json
```diff
@@ -8,7 +8,7 @@
     "dev": "concurrently -n \"server,app,proxy\" \"yarn workspace @template/server dev\" \"yarn workspace @template/app dev\" \"yarn dev:proxy\"",
     "dev:proxy": "node infrastructure/dev/devProxy.js",
     "tsc": "yarn workspaces run tsc",
-    "lint": "eslint 'packages/**/*.ts*' --ignore-path .gitignore --fix",
```
dajinchu, line 11: No more autofix. If you want to autofix, you must `yarn lint --fix`

packages/common/index.ts
```diff
@@ -276,13 +276,13 @@ export type GetQuestionResponse = Question;
...
-  text: string;
+  text!: string;
 
   @IsEnum(QuestionType)
-  questionType: QuestionType;
+  questionType!: QuestionType;
 
   @IsInt()
-  queueId: number;
+  queueId!: number;
```
liustanley, line 285: Just curious, what is the ! for?

NEUDitao: tells typescript to not freak out that the value might be null able, cause the value is never nullable

PR #122, opened 2020-07-23
Clean up queue

### LINKED ISSUE #110: Queue clean up cron job
Clean up questions at the end of the queue, once every hour, or whenever someone checks out

@dajinchu not sure if this should be specified further

willstenzel:
This PR add a function that cleans up a queue when there are no TA's left in the queue. The function is called whenever a TA checks out and every hour on a cron job. 

Closes #110


packages/server/src/course/course.module.ts
```diff
@@ -1,8 +1,9 @@
 import { Module } from '@nestjs/common';
-import { TypeOrmModule } from '@nestjs/typeorm';
 import { CourseController } from './course.controller';
+import { QueueModule } from '../queue/queue.module';
 
 @Module({
   controllers: [CourseController],
+  imports: [QueueModule],
```
willstenzel, line 7: I felt that the cleanup function was better suited as service in the `QueueModule` but it could have been done in the `CouresModule` and not needed to have been imported.

willstenzel:
> Nice!! Seems to work great, just have one thing. When I check out as a TA in a queue with one question, the question status updates to Stale, but it still shows in the queue. I'm not sure if this logic is in the getQuestion query or somewhere on the frontend, but could this be added to the PR?

@liustanley I'm glad you noticed it wasn't working quite right. There is a slight issue in my code I'm still figuring out. For some reason the new `Stale` enum status is not being recognized so when we query for the questions using the `Not(In(Object.values(ClosedQuestionStatus)))` the the `Stale` question status is not being included in there.

willstenzel:
Note: 
@dajinchu and I were working together and decided that the `officehours` and `queue` model should have a relation so that's in here too  :)

packages/server/src/course/course.controller.ts
```diff
@@ -18,52 +18,54 @@ import { User } from '../profile/user.decorator';
...
-      },
-      relations: ['staffList'],
-    });
-
-    const nonEmptyQueues = queues.filter((e) => e.staffList.length > 0);
-
-    const queuesHappeningNow = [];
+    // Set all the queues which should be "Open"
+    course.queues = course.queues.filter((queue) => {
+      const areStaffPresent = queue.staffList.length > 0;
```
dajinchu, line 49: should we just filter on `queue.isOpen`?

willstenzel: yeah good call. Just fixed that up

PR #128, opened 2020-07-25
Student queue improvements

NEUDitao:
image


packages/app/components/Queue/StudentQueueList.tsx
```diff
@@ -173,10 +180,12 @@ export default function StudentQueueList({
           <HeaderText>your question</HeaderText>
         </StudentHeaderCard>
         <EditableQuestion
-          position={3}
+          position={questions.indexOf(studentQuestion) + 1}
```
willstenzel, line 183: Here I think you might want the `?` like you've added on line 322
```suggestion
          position={questions?.indexOf(studentQuestion) + 1}
```

PR #134, opened 2020-07-26
queue improvements

NEUDitao:
image
image


packages/server/src/queue/queue.entity.ts
```diff
@@ -50,6 +52,8 @@ export class QueueModel extends BaseEntity {
   @Expose()
   get queueSize(): number {
     if (!this.questions) {
+      // if you're getting this, make sure you're loading `questions` in relations when you're getting a queue
```
willstenzel, line 55: If the questions are being eager loaded will we no longer run into this issue? 
It's still good to have, I'm just curious.

PR #135, opened 2020-07-26
Remove queue time

willstenzel:
We decided that for now we won't be having a notion of "time" for a queue because it was a bit complicated to get right. It could be implemented later so I added a comment with the commit hash so we can reference back to it.


willstenzel:
> f. one note though, can you link a git_hub_ commit? this way the person doesn't have to jump thru a buncha hoops to find the thing they just click and go "brrr"

I figured we could just add the commit to the commit github url like so 
https://github.com/sandboxnu/office-hours/commit/995e82991587b2077d342b1df87a2665a21c3492

PR #136, opened 2020-07-26
Load in question for queue model so tests pass

willstenzel:
Some tests were erroring with the message `Error: Questions weren't loaded when trying to grab queue size` so this PR fixes that.


dajinchu:
#134 eager loads so we don't need this right?

willstenzel:
> #134 eager loads so we don't need this right?

yup

PR #138, opened 2020-07-26
Navbar Dropdown + Selected

dajinchu:
Navbar highlights current tab. Kind of crappy course selection dropdown.


packages/app/components/Nav/NavBar.tsx
```diff
@@ -119,16 +125,55 @@ export default function NavBar({ courseId }: NavBarProps): ReactElement {
...
+    });
+  }
+
   return (
     <Nav>
       <LogoContainer>
-        {course && (
-          <Logo href={`/course/${courseId}/today`}>{course.name}</Logo>
-        )}
+        <SimpleDropdown overlay={courseSelector}>
```
willstenzel, line 163: Since the majority of students will only have one class they are using this office hours app with, should we only render the dropdown if they have more than one course?

packages/app/components/Nav/NavBar.tsx
```diff
@@ -119,16 +125,55 @@ export default function NavBar({ courseId }: NavBarProps): ReactElement {
     setVisible(false);
   };
 
+  const courseSelector = (
+    <Menu>
+      {profile?.courses.map((c) => (
+        <Menu.Item key={c.course.id}>
+          <Link href="/course/[cid]/today" as={`/course/${c.course.id}/today`}>
```
NEUDitao, line 132: anyone know a guy who can do dropdowns? @zefwang

packages/app/components/Nav/NavBar.tsx
NEUDitao: agree

packages/app/components/Nav/NavBar.tsx
zefwang: I think you're looking for @amielmon. Alternatively, copy the code :)

PR #139, opened 2020-07-26
refactored useQueue and useQuestions

NEUDitao:


packages/app/hooks/useQuestions.ts
```diff
@@ -0,0 +1,33 @@
...
+  const {
+    data: questions,
+    error: questionsError,
+    mutate: mutateQuestions,
+  } = useSWR(
+    qid && `/api/v1/queues/${qid}/questions`,
+    async () => API.questions.index(Number(qid)),
+    {
+      refreshInterval: Math.floor(
+        Math.random() * (FIFTEEN_SECOND_IN_MS - TEN_SECONDS_IN_MS + 1) +
```
willstenzel, line 26: Could we simplify this math to 
`Math.random() * (FIVE_SECONDS_IN_MS + 1) + TEN_SECONDS_IN_MS `

packages/app/hooks/useQueue.ts
```diff
@@ -0,0 +1,32 @@
...
+  queuesError: queueResponse["error"];
+  mutateQueue: queueResponse["mutate"];
+}
+
+export function useQueue(qid: number): UseQueueReturn {
+  const { data: queue, error: queuesError, mutate: mutateQueue } = useSWR(
+    qid && `/api/v1/queues/${qid}`,
+    async () => API.queues.get(Number(qid)),
+    {
+      refreshInterval: Math.floor(
```
willstenzel, line 21: Could someone explain why the `refreshInterval` is needed here but not before?

packages/app/hooks/useQuestions.ts
dajinchu: i actually like Eddy's version more, since it preseves the shape of the interval formula, though it could probably also all be pulled out into a tiny util function. 

Let's not bikeshed over this tho. Probably fine to leave as is

packages/app/hooks/useQueue.ts
dajinchu: The queue is what we want to be polling constantly since it's dynamic. Questions aren't really updated by anyone but the student, so we probably don't need to refresh them.

packages/app/hooks/useQuestions.ts
willstenzel: ohh I see it now. Sounds good!

packages/app/hooks/useQueue.ts
willstenzel: Gotcha. 
I thought it had something to do with it being put in the hook but that clears it up

PR #141, opened 2020-07-26
help next button

NEUDitao:


packages/app/components/Queue/TAQueueList.tsx
```diff
@@ -200,7 +203,13 @@ export default function TAQueueList({
 
   const isStaffCheckedIn = queue?.staffList.some((e) => e.id === user.id);
 
-  const helpNext = async () => {};
+  const helpNext = async () => {
+    const nextQuestion = questions.find(
+      (question) => question.status === QuestionStatusKeys.Queued
```
liustanley, line 208: Is `questions` always strictly sorted by time? If not, this might not work

willstenzel: They are being sorted using the `createdAt: 'ASC'` in the queue.controller

https://github.com/sandboxnu/office-hours/blob/5cd47ddab2f9eba4a6c1a3bc36e5b8cec38ee7bd/packages/server/src/queue/queue.controller.ts#L56-L68

PR #144, opened 2020-07-27
Admin Panel

dajinchu:
admin panel


packages/server/README.md
```diff
@@ -0,0 +1,20 @@
+## How to add a new route
+
+1. Add a method to the controller with `@Get` or other http method decorator. To get request body, path params, query params, etc, use decorators on the arguments. [Available decorators](https://docs.nestjs.com/controllers#request-object). Additionally, we have custom decorators `@User` and `@UserId` to get the currently logged in user.
+2. If a new controller is needed, use `yarn nest g controller <CONTROLLERNAME>` to auto-generate one. 
+3. **IMPORTANT** If route/controller is authenticated: `@UseGuards(JwtAuthGuard)`
+4. If you want to use database model serialization (`@Expose @Exclude` stuff): `@UseInterceptors(ClassSerializerInterceptor)`
+
+## How to access the admin panel
```
dajinchu, line 8: Please follow steps if you want to test locally

packages/server/src/admin/admin-entities.ts
```diff
@@ -0,0 +1,19 @@
+import { AdminEntity } from 'nestjs-admin';
+import { CourseModel } from 'course/course.entity';
+import { QueueModel } from 'queue/queue.entity';
+import { UserModel } from 'profile/user.entity';
+
+export class CourseAdmin extends AdminEntity {
```
dajinchu, line 6: These classes basically tell nestjs-admin how to display the entity. They are optional, as you can also just register the plain typeorm model, but it's nice to show more than just the id in the list. 
We can further customize which fields are shown and which can be searched: https://nestjs-admin.com/docs/admin-entity

packages/server/src/admin/admin.module.ts
```diff
@@ -0,0 +1,29 @@
...
+});
+
+@Module({
+  imports: [CoreModule, AuthModule],
+  exports: [CoreModule, AuthModule],
+})
+export class AdminModule {
+
+  constructor(private readonly adminSite: DefaultAdminSite) {
+    adminSite.register('Course', CourseAdmin);
```
dajinchu, line 25: here's where we could register more entities. I didn't think it was worth adding panels for questions etc.

packages/server/src/admin/admin-user.entity.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Admin users are totally separate from regular users and can only be created from command line.
+ * `yarn cli admin:create`
+ */
+@Entity('admin_user_model')
+export class AdminUserModel extends BaseEntity {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  setPassword(password: string): void {
+    this.passwordHash = hashSync(password, 5);
```
dajinchu, line 14: Please note that we only store the password hash, cause we are `SECURE`

packages/server/src/admin/admin-entities.ts
```diff
@@ -0,0 +1,19 @@
...
+  entity = CourseModel;
+  listDisplay = ['id', 'name'];
+}
+
+export class QueueAdmin extends AdminEntity {
+  entity = QueueModel;
+  listDisplay = ['id', 'room', 'courseId'];
+}
+
```
willstenzel, line 15: Should we add the OfficeHours entity here? Just incase they get switched up last minute on the ical and we haven't re-parsed

packages/server/src/app.module.ts
```diff
@@ -1,44 +1,19 @@
...
-          QuestionModel,
-          QueueModel,
-          DesktopNotifModel,
-          PhoneNotifModel,
-        ],
-        synchronize: true,
-        keepConnectionAlive: true,
-      }),
-    }),
+    TypeOrmModule.forRoot(typeormConfig),
```
willstenzel, line 16: Thanks for cleaning this up  😄

packages/server/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+1. `yarn cli create:admin <username>` then add your password through stdin.
```
liustanley, line 10: We should add that you have to execute this command from within `/packages/server`, Will and I both are dumbo

Also when i did `yarn cli create` before starting the server the user wasn't added, so maybe add a bit saying that you have to start the app first?

packages/server/src/admin/admin.module.ts
liustanley: We might want to add OfficeHour entities here for testing

packages/server/src/admin/admin-entities.ts
dajinchu: going to find some way to add a button to trigger reparse, but later ticket

packages/server/README.md
dajinchu: I will add a top level package json script too that'll work the way `dev:db:seed` does

dajinchu: if DB is up, it should work. server should not be needed

PR #145, opened 2020-07-27
Seeding Buttons

### LINKED ISSUE #126: Testing panel for demo
Different seed scenarios, maybe something similar to the factories

willstenzel:
Adds one button to delete data and another to seed the data to the demo panel on the admin page. 
**Note**: The seed data button also deletes the data before seeding. 

If there are any modifications or additional buttons we need, lmk!

We'll still need to figure out how to conditionally allow this code to build using an evn so we don't expose any of this on prod. 

Closes #126


packages/server/test/util/factories.ts
```diff
@@ -54,6 +54,6 @@ export const QuestionFactory = new Factory(QuestionModel)
   .sequence('text', (i) => `question ${i}`)
   .attr('status', 'Queued')
   .attr('questionType', QuestionType.Other)
-  .attr('createdAt', new Date('2020-03-01T05:00:00.000Z'))
```
dajinchu, line 57: this breaks some integration tests. We can either put this back, or pass `createdAt` to the factories in the failing specs (might just be queue.integration)

willstenzel: Good catch. I added passed a `createdAt` time to the factory to fix the issue

PR #147, opened 2020-07-28
ical scraper cli

dajinchu:
`yarn cli ical:scrape <courseId>` to test. The fixture has the ical url from fundies 1


packages/server/src/app.module.ts
```diff
@@ -11,6 +11,7 @@ import { AdminModule } from './admin/admin.module';
 import { CommandModule } from 'nestjs-command';
 import * as typeormConfig from '../ormconfig';
 import { AdminCommand } from 'admin/admin.command';
+import { ICalCommand } from 'course/ical.command';
```
willstenzel, line 14: I might be missing something but it doesn't like this import, or the one above it is being used in this file.

dajinchu: good catch

packages/server/src/course/course.module.ts
```diff
@@ -1,10 +1,12 @@
...
 import { CourseController } from './course.controller';
 import { QueueModule } from '../queue/queue.module';
-import { AdminModule } from 'admin/admin.module';
+import { ICalCommand } from './ical.command';
+import { IcalService } from './ical.service';
 
 @Module({
   controllers: [CourseController],
   imports: [QueueModule],
+  providers: [ICalCommand, IcalService]
```
liustanley, line 10: can we make this ICalService just to be consistent?

PR #154, opened 2020-07-31
Khoury Login

willstenzel:
* Added new endpoint to receive user data from Khoury
* Refactored some of the login code out of the profile and into the `login.controller`
* Created new abstraction for the testUtils (shoutout @dajinchu )
* Updated nest.js version


packages/server/src/login/login.controller.ts
```diff
@@ -0,0 +1,140 @@
...
+    const payload = (await this.jwtService.decode(token)) as { userId: number };
+    const authToken = await this.jwtService.signAsync({
+      userId: payload.userId,
+    });
+
+    res.cookie('auth_token', authToken).redirect(302, '/');
+  }
+
+  // This is for login on development only
+  @Get('/login/dev')
```
willstenzel, line 131: Decided to separate the production and dev 'entry' endpoints here for the sake of simplicity but it could be done with one

packages/server/src/login/login.controller.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    user = Object.assign(user, {
+      username: body.username,
+      email: body.email,
+      name: body.first_name + body.last_name,
+      photoURL: body.photo_url, // We'll have to find away around this becuase the photo urls expire
+    });
+    await user.save();
+
+    async function courseNameToUserCourse(
```
dajinchu, line 55: this is an N+1. If someone is in 5 courses, this will endpoint will make ~25 database queries. I am fine leaving it in for now, but could you add a TODO? I am looking into application performance monitoring solutions so we can track the performance of each endpoint

willstenzel: Good call. Added that TODO

willstenzel: Good call. I added that TODO

packages/common/index.ts
```diff
@@ -238,6 +239,47 @@ export type PhoneNotifBody = {
...
+  @IsOptional()
+  @IsString()
+  photo_url!: string;
+  
+  @IsOptional()
+  @IsDefined()  // TODO: use ValidateNested instead, for some reason it's crunked
+  courses!: KouryAdminCourse[];
+  
+  @IsOptional()
+  @IsDefined()  // TODO: use ValidateNested instead, for some reason it's crunked
```
NEUDitao, line 264: what's crunked?

packages/server/src/course/course.entity.ts
```diff
@@ -58,4 +58,7 @@ export class CourseModel extends BaseEntity {
   @Exclude()
   // TODO: can we make these not nullable and work with TypeORM
   semesterId: number;
+
+  @Column('boolean', { nullable: true })
+  enabled: boolean;
```
NEUDitao, line 63: what's being enabled here? could do with a better name..

packages/common/index.ts
willstenzel: When I try to use the `@ValidateNested` decorator it can't validate the nested `KhouryAdminCourse` class

packages/server/src/course/course.entity.ts
willstenzel: This flag sets wether or not a course is using the app. Maybe a rename to `isUsingApp` might be better?

willstenzel: I just added a comment to make it more clear

PR #155, opened 2020-08-02
Improve Performance [WIP]

### LINKED ISSUE #96: Load testing & improve performance for production
https://artillery.io/

Find realistic estimate on how many people we expect (worst case scenario maybe 500)
Revalidate rate (maybe 5 seconds, etc....)
100 requests a second?

if we want to we can use AWS to really up the load, prob not necessary

dajinchu:
Adding artillery.io config to load test staging.

Fixes so far:
- `GET /course/:id` Loads raw office hours instead of full entities.


infrastructure/perf/artillery.yml
```diff
@@ -0,0 +1,14 @@
+config:
+  # target: 'http://localhost:3000'
```
dajinchu, line 2: This is a config file for artillery.io which is useful, though I have found https://github.com/mcollina/autocannon to be better at quickly deriving the upper limit of our request/s

infrastructure/prod/nginx.conf
```diff
@@ -47,12 +47,12 @@ http {
 
         gzip on;
 
-        # gzip_vary on;
-        # gzip_proxied any;
-        # gzip_comp_level 6;
-        # gzip_buffers 16 8k;
-        # gzip_http_version 1.1;
-        # gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
+        gzip_vary on;
```
dajinchu, line 50: gzipping with nginx reduced JSON payload by as much as 3x (maybe even more)

packages/server/src/login/login.controller.ts
```diff
@@ -143,6 +148,11 @@ export class LoginController {
     @Query('userId') userId: number,
   ): Promise<void> {
     const token = await this.jwtService.signAsync({ userId });
-    res.cookie('auth_token', token).redirect(302, '/');
+    const isSecure = this.configService
+      .get<string>('DOMAIN')
+      .startsWith('https://');
+    res
+      .cookie('auth_token', token, { httpOnly: true, secure: isSecure })
```
dajinchu, line 155: Snuck this in: the `secure` attributes means this cookie can only be sent over https.

infrastructure/perf/artillery.yml
```diff
@@ -0,0 +1,14 @@
...
+  target: "https://staging.khouryofficehours.com"
+  phases:
+    - duration: 60
+      arrivalRate: 30
+scenarios:
+  - flow:
```
willstenzel, line 8: The other endpoint that might be worth testing is a POST to the `/api/v1/khoury_login` since it does a lot of db queries

infrastructure/perf/artillery.yml
```diff
@@ -0,0 +1,14 @@
...
+      - get:
+          url: "/api/v1/profile/entry?userId=1"
```
willstenzel, line 10: This was route was changed to 
```suggestion
          url: "/api/v1/login/dev?userId=1"
```

willstenzel (approved):
Had a few comment but overall looks solid. Thanks for taking the time to test/improve performance

liustanley (approved):

PR #156, opened 2020-08-02
Notification Registration

dajinchu:
- notification gets own modal
- auto register service worker with next-offline
- prevent duplicate desktop-notifs in database
- register as current user not as hard-coded USER_ID

This flow triggers on the first question they ask. The notification modal is always still accessible in the settings popover.
notif_flow


packages/app/components/Nav/NotificationSettingsModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+            <Switch
+              onChange={async (checked) => {
+                if (checked) {
+                  const canNotify = await requestNotificationPermission();
+                  if (!canNotify) {
+                    form.setFieldsValue({ desktopNotifsEnabled: false });
+                    form.setFields([
+                      {
+                        name: "desktopNotifsEnabled",
+                        errors: ["Please allow notifications in this browser"],
```
NEUDitao, line 84: can we have something more specific ish? Safari users won't intuitively know that they can't subscribe to push notifications.

NEUDitao: On another note, we should tell users to look out for a test notification, and to troubleshoot common things, like the firefox win10 syscall thing, and also turning on their notifications (not that @liustanley or @bugsalexander have ever done that), etc. Desktop notifications can have so many things go wrong 哎呀

packages/app/components/Nav/Settings.tsx
```diff
@@ -1,161 +1,64 @@
...
-              name="desktopNotifsEnabled"
-              valuePropName="checked"
+      <NotificationSettingsModal
+        visible={isNotifOpen}
+        onClose={() => setIsNotifOpen(false)}
+      />
+      <NoPaddingPopover
+        content={
+          <Menu mode="inline">
+            <Menu.Item
```
liustanley, line 36: Small UI thing, but the "Notification Settings" menu item stays highlighted after you click on it, and even after closing

packages/app/components/Queue/StudentQueueList.tsx
```diff
@@ -212,8 +213,28 @@ export default function StudentQueueList({
       setHasDraftInProgress(false);
       finishQuestion(text, qt, isOnline, location);
       closeEditModal();
+      if (isFirstQuestion) {
+        notification.warn({
```
liustanley, line 217: It would be nice if this notification had a mouse hand to make it clear that you can click on it. Like `cursor: pointer`

liustanley:
Another small thing, I'm getting these logs in my console now which don't seem very important, maybe should hide them?

Screen Shot 2020-08-04 at 6 27 26 PM

packages/app/components/Nav/NotificationSettingsModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+interface NotificationSettingsModalProps {
+  visible: boolean;
+  onClose: () => void;
+}
+
+export function NotificationSettingsModal({
+  visible,
+  onClose,
+}: NotificationSettingsModalProps): ReactElement {
+  const { data: profile, error, mutate } = useSWR(`api/v1/profile`, async () =>
```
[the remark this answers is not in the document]
dajinchu: this is the one place we need to mutate, so i figured no

PR #157, opened 2020-08-02
Notify Top of Queue

dajinchu:
This PR notifies when you are third in the queue and when a TA clicks "Help" on your card.


packages/server/package.json
```diff
@@ -95,6 +95,10 @@
       "^.+\\.(t|j)s$": "ts-jest"
     },
     "coverageDirectory": "../coverage",
-    "testEnvironment": "node"
+    "testEnvironment": "node",
+    "moduleDirectories": [
+      "node_modules",
+      "src"
```
dajinchu, line 101: I *think* this might fix the issues with the auto imports, so stuff like
```
import {UserModel} from 'profile/user.entity'
```
is all good now (don't have to do `../../`). Haven't tested it in prod tho.

packages/server/src/notification/notification.service.ts
```diff
@@ -6,19 +6,26 @@ import { UserModel } from '../profile/user.entity';
...
-    'Please respond with either YES or NO. Text STOP at any time to stop receiving text messages',
-  COULD_NOT_FIND_NUMBER:
-    'Could not find an Office Hours account with your phone number.',
-  UNREGISTER:
-    "You've unregistered from text notifications for Khoury Office Hours. Feel free to re-register any time through the website",
-  DUPLICATE:
-    "You've already been verified to receive text notifications from Khoury Office Hours!",
-  OK:
-    'Thank you for verifying your number with Khoury Office Hours! You are now signed up for text notifications!',
+export const NotifMsgs = {
```
dajinchu, line 10: Put all Notification related copy in one spot! Maybe can move this to it's own `notification.constants.ts` file eventually 🤷🏼

packages/server/src/question/question.entity.ts
```diff
@@ -65,4 +71,16 @@ export class QuestionModel extends BaseEntity {
 
   @Column({ nullable: true })
   isOnline: boolean;
+
+  /**
+   * Scopes
+   */
+  static openInQueue(queueId: number): SelectQueryBuilder<QuestionModel> {
```
dajinchu, line 78: Inspired by Rails ActiveRecord scopes. This function gives you a QueryBuilder that has filters applied to give you questions that are open in a certain queue, so you can then join more stuff in, or just get the top 3, etc. If we make our queue order system more complex eventually (to allow manual reordering by TA's), we can just add that logic here!

packages/server/test/util/testUtils.ts
```diff
@@ -82,3 +85,15 @@ export function setupIntegrationTest(
...
 }
+
+const notifMock = mocked({ notifyUser: jest.fn() }, true);
+/**
+ * Module Modifier tests can pass to setupIntegrationTest to mock the notifService and to expect things
+ * ex:
+ * const supertest = setupIntegrationTest(QuestionModule, modifyMockNotifs);
+ */
+
+export const modifyMockNotifs: ModuleModifier = (t) =>
```
dajinchu, line 96: ngl really proud of this code lmao. 
See the question integration test for usage. It let's us check whether a user has been notified without being a huge pain

packages/server/src/question/question.subscriber.ts
```diff
@@ -0,0 +1,45 @@
...
+    connection.subscribers.push(this);
+  }
+
+  listenTo() {
+    return QuestionModel;
+  }
+
+  // Notify students when they get near the top of the queue
+  async afterUpdate(event: UpdateEvent<QuestionModel>) {
+    // if status updated to closed
```
willstenzel, line 29: Looks like this is triggered when any question's status is set to a `ClosedQuestionStatus`. Is it possible that if a question below the first three questions is changed to a `ClosedQuestionStatus` then the person in third place will get notified multiple times? (e.g when a student deletes their question or the TA removes it from the queue)

willstenzel (approved):
Had some comments but overall looks good!

Good call using the `EventSubscriber` and I liked how you figured out how to cleanly test the notification stuff

packages/server/src/question/question.subscriber.ts
dajinchu: good catch. Crap. not sure how to fix

packages/server/src/question/question.controller.ts
```diff
@@ -214,7 +224,7 @@ export class QuestionController {
 
     this.notifService.notifyUser(
```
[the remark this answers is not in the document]
dajinchu: gotta await or return it. fixed.

packages/server/src/question/question.subscriber.ts
dajinchu: fixed now by checking that the student wasn't already third in the queue

NEUDitao: something to keep in mind is that if there are three TAs helping students, and we're only checking for open questions to notify, no one's gonna get notified.

NEUDitao (approved):

PR #167, opened 2020-08-04
Move dev panel to separate route and update login page

### LINKED ISSUE #151: New login page & move dev panel to separate route
**Description:**
Our new `/login` page will be a single button that users can click to redirect to the Khoury Admin login flow. Since our current `/login` page is really a dev panel, we should move it to a separate route.

**Specs:**
- [ ] Current `/login` page is moved to `/devpanel` route
- [ ] New `/login` page contains a button displaying "Login with Khoury Admin"
- [ ] New login button integrates with #154 (can just branch off of this)

If blocked, contact @willstenzel

willstenzel:
Closes #151 

@liustanley I did what I could, but feel free to update the styling on the login page.


packages/app/pages/dev.tsx
```diff
@@ -0,0 +1,120 @@
...
+`;
+
+const PageHeader = styled.div`
+  display: flex;
+  justify-content: center;
+  margin-top: 40px;
+`;
+
+export const getStaticProps: GetStaticProps = async () => {
+  const hidePage = (PROD_URL === process.env.DOMAIN);
```
willstenzel, line 49: @dajinchu will the `DOMAIN` env be set on the production server?

dajinchu: yes, but it's not available in the `packages/app` project... I think I'm going to redo how we handle environment variables

willstenzel:
Should I make the logout button automatically route to `/dev` instead of `/logout` so we keep the current dev experience the same?

PR #169, opened 2020-08-05
Ta check in tests

NEUDitao:


willstenzel:
Just pushed a commit to update the seeds so that the test to check in `"from the today page with a room already existing"` can run without the use of `cy.mock`. 
This is needed sine we'll be running all the test against the real API in the CI

willstenzel:
I'm going to go ahead and merge this in so I can rebase off of it in my PR

PR #171, opened 2020-08-06
fixed seeding && fixed waittime on TA's

NEUDitao:
gentle reminder that any change you make in TAQueueCard and TAQueueList must be done in student as well and vice versa


packages/app/components/Queue/TAQueueCard.tsx
```diff
@@ -23,6 +23,7 @@ export default function TAQueueCard({
   question,
   onOpen,
 }: TAQueueCardProps): ReactElement {
+  const waitTime = now.getMinutes() - question.createdAt.getMinutes();
```
liustanley, line 26: i think you forgot to copy over `now`

willstenzel:
```suggestion
  const now = new Date();
  const waitTime = now.getMinutes() - question.createdAt.getMinutes();
```

NEUDitao: i did and then committed but never pushed -- there's a new bug there you might wanna check out :P

packages/server/src/question/question.entity.ts
```diff
@@ -45,7 +45,7 @@ export class QuestionModel extends BaseEntity {
   @Exclude()
   taHelpedId: number;
 
-  @CreateDateColumn()
+  @Column()
```
dajinchu, line 48: What's the reason for this?

liustanley: CreateDateColumn wasn't correctly storing the createdAt date, it did something funky with the UTC conversion and actually messed up our times. I changed it to a normal Column and pass in `new Date()` when creating a question which works fine, but not sure if we should investigate this further.

liustanley:
https://github.com/typeorm/typeorm/issues/3286

PR #175, opened 2020-08-07
Elastic APM

dajinchu:
Most of the changes are actually just prettier....

All this does is install the Elastic APM agent on the backend and frontend so timings and errors get reported to Elastic Cloud. 

Eventually we might also want to send logs and NGINX+Postgres stats using Filebeat, but not atm.


packages/server/elastic-apm-node.js
```diff
@@ -0,0 +1,8 @@
+require('dotenv').config();
+
+module.exports = {
+  serviceName: new URL(process.env.DOMAIN).hostname.replace(/\./g, '-'),
+  secretToken: process.env.ELASTIC_APM_SECRET_TOKEN,
+  serverUrl: process.env.ELASTIC_APM_SERVER_URL,
+  active: process.env.NODE_ENV === 'production',
```
dajinchu, line 7: Only run in staging and production. It is mostly meaningless to run in dev.

packages/server/src/apm.interceptor.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as apm from 'elastic-apm-node';
+
+@Injectable()
+export class ApmInterceptor implements NestInterceptor {
+  intercept(
+    context: ExecutionContext,
+    next: CallHandler,
+  ): Observable<Response> {
+    return next.handle().pipe(
+      catchError((error) => {
```
dajinchu, line 19: This sends all errors over to APM, so we can be notified when things go 💥

PR #180, opened 2020-08-08
Cypress github action

willstenzel:


.github/workflows/cypress.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: E2E Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        with:
+          record: true
+          start: yarn dev
```
willstenzel, line 16: @dajinchu should we be using the production or development build here?

dajinchu: probably production so that we're less prone to timeouts. This means we need to do first `yarn build` then `yarn dev:proxy` and `yarn prod:start` although i'm not sure `yarn prod:start` will work on GitHub causes it uses pm2 to run in the background. lmk what happens.

PR #182, opened 2020-08-09
Protect Get Endpoints

aislinblack:
We created a guard to verify user roles within a course.


packages/server/src/question/question-role.guard.ts
```diff
@@ -0,0 +1,42 @@
...
+import { QueueModel } from '../queue/queue.entity';
+import { RolesGuard } from '../guards/role.guard';
+
+@Injectable()
+export class QuestionRolesGuard extends RolesGuard {
+  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+  async setupData(
+    request: any,
+  ): Promise<{ courseId: number; user: UserModel }> {
+    let queueId = request.params.queueId;
```
dajinchu, line 17: If QuestionRolesGuard is only for question controller, in what cause is there actually a `queueId` in the params?

aislinblack: This is specifically for the create question endpoint. (if I remember correctly) Because in that case we don't have a preexisting question id, but we do have the queueid

packages/server/test/question.integration.ts
```diff
@@ -39,7 +53,8 @@ describe('Question Integration', () => {
 
   describe('POST /questions', () => {
     it('posts a new question', async () => {
-      const queue = await QueueFactory.create();
+      const course = await CourseFactory.create();
+      const queue = await QueueFactory.create({ courseId: course.id });
```
willstenzel, line 57: Just curious why we need this change?

aislinblack: The problem with the guard is that then you need to explicitly have a queue and course for it to run successfully which is true when you're running the app, but not in this test.

packages/app/hooks/useRoleInCourse.ts
```diff
@@ -3,5 +3,5 @@ import { useProfile } from "./useProfile";
 
 export function useRoleInCourse(courseId: number): Role {
   const profile = useProfile();
-  return profile?.courses.find((e) => e.course.id === courseId).role;
+  return profile?.courses.find((e) => e.course.id === courseId)?.role;
```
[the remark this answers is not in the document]
aislinblack: Yeah I'm not really sure (I put in that change a long while ago lol) but I think I largely saw it error when our endpoint returned an error message, but I'm really not sure why looking at the logic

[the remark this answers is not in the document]
willstenzel: Okay, I think we're in the same boat. 

I wonder if there is a deeper issue that might be causing the error. @dajinchu any thoughts on this?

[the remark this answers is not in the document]
dajinchu: dunno, maybe when you are the page for a course you are not in somehow

packages/server/src/question/question-role.guard.ts
dajinchu: Ah i see. might be better to just do the authorization directly in the endpoint for something like that then

packages/app/hooks/useRoleInCourse.ts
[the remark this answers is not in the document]
willstenzel: Ahh yeah I think that's it. Should we add an issue to redirect when the user visits a page for a course they're not in?

packages/server/src/question/question-role.guard.ts
aislinblack: I wanted I wanted to keep it because there was alot of repeated code within each of the endpoints which is stopped by the guard, but definitely something to consider.

PR #183, opened 2020-08-09
Show office hour times in queue/today page

### LINKED ISSUE #161: Show office hour times in queue/today page
_Dev design required_

**Description:**
Office hour times are currently hardcoded in the `/queue` page and not shown in the `/today` page. We'd like to show these time ranges (e.g. 2:00pm - 4:00pm) generated from our objects.

**Specs:**
- [x] Office hour time correctly shows in the following format: `2:00pm - 4:00pm`
- [x] Time is shown in the top right corner of an today page queue card
- [x] Time is shown in the queue page header (currently hardcoded)

Contact @dajinchu for dev design, or if blocked.

liustanley:
Derives queue times from existing office hour objects

Closes #161


willstenzel:
I thought adding times to the queue was too complicated but you guys proved me wrong!

packages/server/src/queue/queue.service.ts
```diff
@@ -1,14 +1,84 @@
...
+interface TimeInterval {
+  startTime: Date;
+  endTime: Date;
+}
 
 @Injectable()
 export class QueueService {
   constructor(private connection: Connection) {}
 
+  public async addQueueTimes(queue: QueueModel) {
```
dajinchu, line 18: Not sure about having this code in `QueueService`. I wonder if it'd make more sense in the QueueModel, since this code is sort of intrinsic to the queue data?

liustanley:
> One minor edge case to think about is what happens if a TA stays past their scheduled office hours or checks in early? Should we look for the closest time interval to the current time and use that for the queue?

@willstenzel I've been thinking about this and I think defaulting to the closest time interval to the current time will result in incorrect times for custom queues. What I'm thinking that might be a better solution would be to look for the closest time interval with +/- 15 minutes to account for most edge cases. We can then make `startTime` and `endTime` optional, so that for custom queues the time just doesn't display.

PR #184, opened 2020-08-10
Server Sent Events

dajinchu:
Server Sent Events lets our server proactively send data to the frontend without the frontend explicitly requesting it. This means: 
- We can update the queue near instantaneously instead of every 10-15 seconds. 
- We might reduce server load by avoiding polling and hitting the database less. (hit DB once and send out the data to everyone INSTEAD of having everyone ask for data every 10 seconds and hitting DB once for each request)

## Demo

Left and right are different students. I don't ALT+TAB or anything, it just updates.

sse


NEUDitao (approved):

willstenzel (approved):
🔥

PR #187, opened 2020-08-11
Calendar stuff

ofloody:
Creating a queue when an office hour block is created from calendar event (iCal)


dajinchu:
Tests should now work if you run `yarn test:unit ical`in `packages/server`

willstenzel:
> i get a JS Heap error after scraping on my computer but it might be because of potato according to @NEUDitao ...
> 
> can someone else verify that scraping doesn't crash everything? otherwise looks great!!

@liustanley scraping doesn't crash everything on my local box so it might be because of the potato issue

dajinchu:
> > i get a JS Heap error after scraping on my computer but it might be because of potato according to @NEUDitao ...
> > can someone else verify that scraping doesn't crash everything? otherwise looks great!!
> 
> @liustanley scraping works on my local box so it might be because of the potato issue

I fixed this issue in 1f45fda

PR #191, opened 2020-08-12
removed student names/pictures, added questions

NEUDitao:
wwwwwwwwwww


liustanley:
Wait a bit before merging this, I was still thinking about how we should display anonymous students in the queue. I'll let you know when I get this finalized.

PR #193, opened 2020-08-12
TA Status Cards

dajinchu:
image


packages/app/components/RenderEvery.tsx
```diff
@@ -0,0 +1,25 @@
+import { useState, useEffect, useCallback, ReactElement } from "react";
+
+interface RenderEveryProps {
+  render: () => ReactElement;
+  interval: number;
+}
+export function RenderEvery({
```
willstenzel, line 7: This is slick, I like it

packages/app/components/Queue/StudentQueueList.tsx
```diff
@@ -299,6 +300,7 @@ export default function StudentQueueList({
                 )}
               </Col>
             </Row>
+            <StatusRow questions={questions} taList={queue.staffList} />
```
willstenzel, line 303: This kinda removes the anonymity added in https://github.com/sandboxnu/office-hours/pull/191

I'm not sure what makes the most sense regarding student anonymity, might be worth having a discussion about.

packages/app/components/Queue/StatusRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+  helpedAt: Date;
+}
+function HelpingFor({ studentName, helpedAt }: HelpingForProps): ReactElement {
+  return (
+    <RenderEvery
+      render={() => (
+        <span>
+          Helping <BlueSpan>{studentName}</BlueSpan> for{" "}
+          <BlueSpan>
+            {Math.round((Date.now() - helpedAt.getTime()) / 60000) + " min"}
```
liustanley, line 115: We've been putting time related calculations into `/util/TimeUtil.ts` if you want to add this into there

packages/app/components/Queue/StudentQueueList.tsx
liustanley: Yeah, how about for TA queues we keep it the same, but for a student queue all they can see is "Helping a student for 23 minutes"

PR #200, opened 2020-08-17
Prevent new questions in closed queues

### LINKED ISSUE #181: Prevent New Questions in Closed Queues
Students are on the queue page when it closes have NO way of knowing it's closed and keep joining it.

- [ ] `POST /questions` should not allow if queue is closed
- [ ] Add `isOpen` to `QueuePartial`
- [ ] Frontend should display a "queue has closed" screen.

liustanley:
Closes #181

### DIFF
--- a/packages/server/src/course/course.controller.ts
+++ b/packages/server/src/course/course.controller.ts
@@ -61,6 +61,7 @@ export class CourseController {
     @Param('room') room: string,
     @User() user: UserModel,
   ): Promise<QueuePartial> {
+    console.log('start of function');
     let queue = await QueueModel.findOne(
       {
         room,
@@ -70,6 +71,7 @@ export class CourseController {
     );
 
     if (!queue) {
+      console.log('!queue');
       queue = await QueueModel.create({
         room,
         courseId,
@@ -80,7 +82,6 @@ export class CourseController {
 
     queue.staffList.push(user);
     await queue.save();
-
     return queue;
   }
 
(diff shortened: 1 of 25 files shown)


packages/server/src/queue/queue.entity.ts
```diff
@@ -60,18 +60,24 @@ export class QueueModel extends BaseEntity {
   startTime: Date;
   endTime: Date;
 
-  async isOpen(): Promise<boolean> {
-    if (this.staffList.length > 0) {
+  isOpen: boolean;
+
+  async checkIsOpen(): Promise<boolean> {
+    console.log(this.staffList);
```
dajinchu, line 66: remember to remove

packages/server/src/course/course.controller.ts
```diff
@@ -61,6 +61,7 @@ export class CourseController {
     @Param('room') room: string,
     @User() user: UserModel,
   ): Promise<QueuePartial> {
+    console.log('start of function');
```
dajinchu, line 64: remember to remove these logs

packages/server/src/queue/queue.service.ts
```diff
@@ -18,6 +18,7 @@ export class QueueService {
     });
 
     await queue.addQueueTimes();
+    await queue.checkIsOpen();
```
willstenzel, line 21: This totally works but it might be slightly more clear to append the `queue.isOpen` attribute after this line like it's done with `queue.questions`, instead of doing in the `queue.checkIsOpen` function

PR #204, opened 2020-08-18
Page For No Courses

dajinchu:


packages/server/src/login/login.controller.ts
```diff
@@ -147,12 +139,24 @@ export class LoginController {
...
-    const token = await this.jwtService.signAsync({ userId });
+    this.enter(res, userId);
+  }
+
+  // Set cookie and redirect to proper page
+  private async enter(res: Response, userId: number) {
+    const userCourse = await UserCourseModel.findOne({
+      where: { userId },
+    });
+    console.log(userCourse);
```
willstenzel, line 150: Do you mean to keep this?

packages/server/src/login/login.controller.ts
```diff
@@ -147,12 +139,24 @@ export class LoginController {
...
   ): Promise<void> {
-    const token = await this.jwtService.signAsync({ userId });
+    this.enter(res, userId);
+  }
+
+  // Set cookie and redirect to proper page
+  private async enter(res: Response, userId: number) {
+    const userCourse = await UserCourseModel.findOne({
+      where: { userId },
+    });
```
willstenzel, line 149: We might want to check that the current semester of the course is the current semester so it the query doesn't return an old course. But for now it should work since it's everyone's first semester

dajinchu: good point, we'll have to go through the whole thing probably to support semesters later

PR #206, opened 2020-08-20
Queue toggle to allow questions 

NEUDitao:
image
image
image
image


packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -51,6 +121,17 @@ export default function QueueListHeader({
           </NotesText>
         </Container>
       )}
+      <div style={{ paddingLeft: "64px" }}>
+        {queue?.allowQuestions ? (
+          <NotesText style={{ color: "green" }}>
+            This queue is allowing new questions
```
willstenzel, line 127: Since students probably assume the queue is allowing new questions we might only need to show the text when it's **not** allowing new questions. 

Either way works tho

packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -22,21 +25,88 @@ const TimeText = styled.div`
...
+        title="LOL Stanley you're gonna have to figure this one out"
+        visible={queueSettingsModal}
+        onCancel={() => {
+          setUpdatedNotes(oldNotes);
+          setQueueSettingsModal(false);
+        }}
+        onOk={() => {
+          setOldNotes(updatedNotes);
+          setQueueSettingsModal(false);
+        }}
```
willstenzel, line 88: Looks like `updateQueueNotes` need to be called somewhere in here
```suggestion
        onOk={() => {
          setOldNotes(updatedNotes);
          updateQueueNotes();
          setQueueSettingsModal(false);
        }}
```

PR #208, opened 2020-08-24
Into popup at first login

NEUDitao:


NEUDitao:
gdi cypress

willstenzel:
I'm not sure it makes sense to be using `localStorage` to determine wether or not to show the popup. Using `localStorage`, the popup would show any time someone's browser cache is cleared or they sign in on a new browser.   

One possibility is to store this info with the user profile on the backend, would love to hear some other opinions tho.
@dajinchu any thoughts on this?

NEUDitao:
> I'm not sure it makes sense to be using `localStorage` to determine wether or not to show the popup. Using `localStorage`, the popup would show any time someone's browser cache is cleared or they sign in on a new browser.
> 
> One possibility is to store this info with the user profile on the backend, would love to hear some other opinions tho.
> @dajinchu any thoughts on this?

other websites do it all the time, like good ol figma or generalsio -- I think it's better this way, daj think sso too

packages/app/components/Today/WelcomeStudents.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const [firstTime, setFirstTime] = useLocalStorage("firstTime", true);
+
+  return (
+    <Modal
+      visible={firstTime}
+      footer={null}
+      onCancel={() => setFirstTime(false)}
+    >
+      <div>
+        Welcome to the Khoury Office Hours app! This project was started by
```
liustanley, line 15: wow can't believe you used a real message omo

packages/app/pages/course/[cid]/today.tsx
```diff
@@ -57,6 +58,7 @@ export default function Today(): ReactElement {
   }
   return (
     <div>
+      <WelcomeStudents />
```
Comment (liustanley) [inline packages/app/pages/course/[cid]/today.tsx:61]: this is small, but it might be better to use a conditional render here so that we don't always render the modal in the DOM. if this messes up any ant design animations then no worries

`{firstTime && <WelcomeStudents />}`

PR #210, opened 2020-08-24
Fix 3rd in Queue Notif Error

dajinchu:
When you go from 3 people in the queue, to 2 people in the queue, the subscriber throws an error because `third` is undefined.


dajinchu:
> didn't we have this and then change it and then change it back????
> 
> in other news, what if `previousThird` is null?

We use the `?` so that's fine. The issue is when we enter the `if` block and destructure `third`

PR #211, opened 2020-08-24
Queue Redesign

### LINKED ISSUE #196: Student 'Helping' page
_Still needs design, @liustanley will update this card when ready for implementation._

**Description:**
When a TA starts helping a student, we want that student's UI to update to reflect this status change. This helping page would display information such as guidelines on what to do for a Teams call, the TA coming to help them, and a link to open the Teams chat.

Contact @liustanley if blocked.

dajinchu:
Closes #196 and #211


packages/app/components/Queue/StudentBanner.tsx
```diff
@@ -0,0 +1,169 @@
...
+          title={
+            <span>
+              <BoldNumber>{studentQuestion.taHelped.name}</BoldNumber> is coming
+              to help you
+            </span>
+          }
+          buttons={
+            <>
+              <LeaveQueueButton leaveQueue={leaveQueue} />
+              <BannerButton icon={<TeamOutlined />}>Open Teams DM</BannerButton>
```
willstenzel, line 139: Looks like this needs an link somewhere for it to open the Teams DM

packages/app/components/Queue/TABanner.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <Avatar
+              size={64}
+              icon={<UserOutlined />}
+              src={helpingQuestion.creator.photoURL}
+            />
+          </Col>
+          <Col>
+            <InfoHeader>question</InfoHeader>
+            <Info>{helpingQuestion.text ?? ""}</Info>
+            <InfoHeader>location</InfoHeader>
```
willstenzel, line 57: Should we also include the questions type somewhere in the banner?

willstenzel:
I'm happy to fix up the cypress tests tomorrow. Looks like the one that is not passing is the `allow_or_disable_questions.spec.js` which is the one I worked added recently so it should be quick

packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -1,126 +1,122 @@
...
+
+export const QueuePageContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  @media (min-width: 767px) {
+    flex-direction: row;
+  }
+`;
+
+export const VerticalDivider = styled.div`
```
liustanley, line 18: If possible, could we get this divider to extend for the full height of the page instead of cutting off at where the content ends?

packages/app/components/Queue/StudentQueueList.tsx
```diff
@@ -218,11 +218,16 @@ export default function StudentQueueList({
           />
           <VerticalDivider />
           <Space direction="vertical" size={40} style={{ flexGrow: 1 }}>
-            <StudentBanner
-              queueId={qid}
-              editQuestion={openEditModal}
-              leaveQueue={leaveQueue}
-            />
+            {studentQuestion &&
```
dajinchu, line 221: @liustanley what's the point of this? Isn't the conditional render already handled inside studentbanner?

liustanley: I added that because `StudentBanner` still returns a DOM object, so the `Space` component adds unnecessary padding above the Queue header when a student question isn't present. I'm realizing now though that I only have to check for studentQuestion so I'll update that.

liustanley:
Closes #196 and #211

PR #213, opened 2020-08-27
Remove fixtures and seed commands

willstenzel:


dajinchu:
I think we need to add to the seed button or the student login button, because now you can't use the `login as student 1` button since userid 1 doesn't exist without the fixture

willstenzel:
> I think we need to add to the seed button or the student login button, because now you can't use the `login as student 1` button since userid 1 doesn't exist without the fixture

Good catch, I'll add that in.

PR #224, opened 2020-08-28
bug fix for join queue popconfirm bug

NEUDitao:
ligma


packages/app/components/Queue/StudentQueueList.tsx
```diff
@@ -130,6 +130,7 @@ export default function StudentQueueList({
         const newQuestions = [...questions, createdQuestion];
         await mutateQuestions(newQuestions);
         setPopupEditQuestion(true);
+        return true;
```
NEUDitao, line 133: this is the only actual change i'm not kidding

dajinchu: dangothy

PR #225, opened 2020-08-29
Fix "Can't Find" flow 

NEUDitao:
Can't Find flow is now happier


packages/app/components/Queue/StudentQueueList.tsx
```diff
@@ -298,3 +319,30 @@ function QueueQuestions({ questions, studentQuestion }: QueueProps) {
...
+      footer={[
+        <Button key="leave" danger onClick={props.leaveQueue}>
+          Leave Queue
+        </Button>,
+        <Button type="primary" key="rejoin" onClick={props.rejoinQueue}>
+          Rejoin Queue
+        </Button>,
+      ]}
+      closable={false}
+      title="哎呀！！"
```
willstenzel, line 342: I like the humor but I think it's getting to be that time where we should avoid adding random stuff so we don't risk missing removing something like this before the semester starts.

willstenzel:
When I tried testing it on my local, when I click "Leave Queue" on the `CantFindModal` I got a popup asking if I wanted to delete my previous question. Is this happening for anyone else?

Screen Shot 2020-08-31 at 12 41 14 AM

NEUDitao:
> When I tried testing it on my local, when I click "Leave Queue" on the `CantFindModal` I got a popup asking if I wanted to delete my previous question. Is this happening for anyone else?
> 
> Screen Shot 2020-08-31 at 12 41 14 AM

fixed in anothe rPR

packages/app/components/Queue/StudentQueueList.tsx
NEUDitao: 哎呀

PR #226, opened 2020-08-29
Update Khoury login endpoint

### LINKED ISSUE #218: Handle student course sections
**Description:**
From the Khoury meeting, we learned that a single course such as Fundies 1 will have three separate office hour sections. In our app, we can treat each section as a distinct "Course", but we need to create a `CourseGroupMapping` table to place students into their respective courses when they login. This table should be populated manually via the admin panel.

More details here, but only handle the student flow for this ticket: https://www.notion.so/Course-Sections-2cab2029bd2d47d68216663146169df8

**Specs:**
- [x] Create `CourseGroupMapping` table with specified columns in notion doc
- [x] Populate `CourseGroupMapping` table and `CourseModels` with the admin panel
- [x] When a student logs in, query `CourseGroupMapping` with student courses/sections to obtain each `CourseModel` available in our database
- [x] Create a `UserCourse` entity for each `CourseModel` found
- [x] Ensure app functions the same

If blocked, contact @liustanley or @dajinchu

willstenzel:
Note: The username is no longer being sent to us by Khoury so the `email` is now the unique identifier for a user

Closes #218

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -251,9 +251,6 @@ export type GetProfileResponse = User;
 
 export class KhouryDataParams {
   @IsString()
-  username!: string;
-
-  @IsString()
   email!: string;
 
   @IsString()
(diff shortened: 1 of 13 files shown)


packages/common/index.ts
```diff
@@ -265,31 +262,45 @@ export class KhouryDataParams {
...
+  @IsInt()
+  crn!: number;
 
   @IsString()
-  semester!: number;
+  course!: string;
 
   @IsBoolean()
-  withdraw!: boolean;
+  acclerated!: boolean;
```
NEUDitao, line 286: `acclerated`

willstenzel: Good catch!

packages/server/src/login/login.controller.ts
```diff
@@ -36,68 +40,55 @@ export class LoginController {
...
     );
+    
     await Promise.all(
-      body.ta_courses
-        .filter((c) => !c.withdraw)
-        .map(async (c) => {
-          const taCourse = await courseNameToUserCourse(c.course_name, Role.TA);
+      body.ta_courses.map(async (c: KhouryTACourse) => { 
+        // Query for all the courses which match the name of the generic course from Khoury
+        const courses = await getRepository(CourseModel)
```
dajinchu, line 78: Would it make sense to use the `CourseSectionMappingModel`?
Like:
``` 
await CourseSectionMappingModel.find({
  where: { genericCourseName: courseName },
  relations: ['course'],
});
```

willstenzel: Absolutely

PR #231, opened 2020-09-01
Remove unnecessary db query

willstenzel:
Looks like we don't need this query. 

Getting questions for the queue comes from the `/queues/:queue_id/questions` endpoint

### DIFF
--- a/packages/server/src/queue/queue.service.ts
+++ b/packages/server/src/queue/queue.service.ts
@@ -19,8 +19,6 @@ export class QueueService {
 
     await queue.addQueueTimes();
     await queue.checkIsOpen();
-    const questions = await QuestionModel.find({ where: { queueId } });
-    queue.questions = questions;
 
     return queue;
   }


NEUDitao (approved):

dajinchu (changes requested):
don't we need questions to be loaded for the `queueSize` property on `QueueModel`? 
https://github.com/sandboxnu/office-hours/blob/5f7268d6ee0f3d52eaadab9616c68a12e7d9ad52/packages/server/src/queue/queue.entity.ts#L81-L85

willstenzel:
> don't we need questions to be loaded for the `queueSize` property on `QueueModel`?
> https://github.com/sandboxnu/office-hours/blob/5f7268d6ee0f3d52eaadab9616c68a12e7d9ad52/packages/server/src/queue/queue.entity.ts#L81-L85


I think they are being eagerly loaded so it's not needed anymore. 
https://github.com/sandboxnu/office-hours/blob/849727aa3208747a584180e4bd8baba0dd05d2fd/packages/server/src/queue/queue.entity.ts#L42-L46

liustanley (approved):

dajinchu (approved):
Looks good. We may eventually need to un-eager load, but let's wait for real production perf stats to come in.

PR #233, opened 2020-09-02
Server Sent Events for the Queue

aislinblack:


packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -105,6 +111,12 @@ export function QueueInfoColumn({
           <QueuePropertyText>{queue.notes}</QueuePropertyText>
         </QueuePropertyRow>
       )}
+      <QueuePropertyRow>
+        {isQueueLive ? <CloudSyncOutlined /> : <FrownOutlined />}
```
aislinblack, line 115: @liustanley we await a better design

NEUDitao: YEAH FROWN OUTLINED I'M LIVING FOR THIS

packages/app/hooks/useQueue.ts
```diff
@@ -1,32 +1,43 @@
...
-    }
+    async () => API.queues.get(Number(qid))
+  );
+
+  const isQueueLive = useEventSource(
+    qid && `/api/v1/queues/${qid}/sse`,
+    "queue",
+    useCallback(
+      (data) => {
+        console.log(data);
```
NEUDitao, line 27: console log

NEUDitao (approved):

packages/app/components/Queue/QueueListSharedComponents.tsx
liustanley: i fixed this in my branch rip but its all good i'm here for frown outlined

liustanley: oh wait just kidding this is something new i think it looks good!

liustanley (approved):
this is awesome!!! code is too chungus brain for me though

packages/server/src/queue/queue.controller.ts
```diff
@@ -49,19 +49,11 @@ export class QueueController {
...
-    if (role === Role.STUDENT) {
-      return questions.map((question) => {
-        if (question.creator.id !== user.id) {
-          question.creator = pick(question.creator, ['id']);
-        }
-        return question;
-      });
-    }
-    return questions;
+    return this.queueService.anonymizeQuestions(questions, userId, role);
```
willstenzel, line 56: I like that you abstracted this but it might be more readable if you left the `if (role === Role.STUDENT) {` check in there so that the `this.queueService.anonymizeQuestions` would only be anonymizing the data it if the student was the one hitting the endpoint. 

It's a minor thing tho, so if you want to leave it that's totally fine.

willstenzel (approved):
I'm in the same boat as Stanley on this one. I'm not familiar with SSE but I can see that it works which is super cool!

If someone wanted to give a quick high level overview of how this code works, I'd love to listen.

PR #241, opened 2020-09-03
Can't find test -> remove from queue test

ofloody:


willstenzel (commented):
This is great but I'm not sure we are going to be using "Can't Find" anymore :/

We don't have a test for what happens when a TA removes a student from the queue, which is a very similar flow, so maybe this test could be re-purposed for that?

liustanley (approved):

liustanley:
> This is great but I'm not sure we are going to be using "Can't Find" anymore :/
> 
> We don't have a test for what happens when a TA removes a student from the queue, which is a very similar flow, so maybe this test could be re-purposed for that?

We're still going to be using "Can't Find", just not for initial rollout. I agree that this would be a good base for a remove from queue test, but don't delete this test! It'll be good for later, we can just disable it temporarily.

willstenzel (approved):

Issue #235, opened 2020-09-03
Remove 'Helping' students from the queue


liustanley:
Not actually remove, but just don't display them in the queue anymore

- [ ] Change the StudentQueueList's `QueueQuestions` to filter out Helping questions
- [ ] Fix useStudent hook

PR #249, opened 2020-09-04
Limbo queue list

ofloody:


dajinchu (approved):
Are we going to get rid of the Can't Find button? Otherwise students will see people cut in line in front of them right?

liustanley (changes requested):
StudentQueueList looks good! For the TAQueueList, we don't need to separate `queuedQuestions` and `helpedQuestions`. I think TA's would be confused by a bunch of 0's at the top of the list, we can leave it all numbered like before.

Another thing is that we also need to adjust for the StudentBanner with the `useStudentQuestion` hook, as mentioned in #235

liustanley (approved):

dajinchu (approved):

PR #253, opened 2020-09-05
Check Khoury webhook for signature

willstenzel:


packages/server/src/login/khoury-public-key.pem
```diff
@@ -0,0 +1,7 @@
+
+-----BEGIN PUBLIC KEY-----
+MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDCFENGw33yGihy92pDjZQhl0C3
+6rPJj+CvfSC8+q28hxA161QFNUd13wuCTUcq0Qd2qsBe/2hFyc2DCJJg0h1L78+6
+Z4UMR7EOcpfdUE9Hf3m/hs+FUR45uBJeDK1HSFHD8bHKD6kv8FPGfJTotc+2xjJw
+oYi+1hqp1fIekaxsyQIDAQAB
+-----END PUBLIC KEY-----
```
willstenzel, line 7: This is a place-holder public key. We'll have to replace with the real one when Khoury hooks it up on their end

packages/server/src/login/http-signature.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from '@nestjs/common';
+import * as fs from "fs";
+import * as httpSignature from 'http-signature';
+import { Request } from 'express';
+
+
+@Injectable()
+export class HttpSignatureService {
+    public verifyRequest(req: Request): boolean {
+        const PUBLIC_KEY = fs.readFileSync(__dirname + '/khoury-public-key.pem', 'ascii');
```
willstenzel, line 10: Not sure if this could be made a class variable so for better performance?

willstenzel:
I'm planning to add a test in the `http-signature.service.spec.ts` before marking it "Ready for Review."

packages/server/src/login/http-signature.service.ts
dajinchu: Yes please do this in the constructor

willstenzel: done

willstenzel:
> I'm planning to add a test in the `http-signature.service.spec.ts` before marking it "Ready for Review."

Actually I'm not so sure the best way to write a test for this. 

In the example [here](https://github.com/joyent/node-http-signature#client) they use another library called `https` but not sure we want to install another dependency just for a test. Any help would be appreciated.

packages/server/src/login/http-signature.service.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Request } from 'express';
+
+
+@Injectable()
+export class HttpSignatureService {
+    PUBLIC_KEY: string;
+    constructor() {
+        this.PUBLIC_KEY = fs.readFileSync(__dirname + '/khoury-public-key.pem', 'ascii');
+    }
+    public verifyRequest(req: Request): boolean {
```
dajinchu, line 13: I think we should just put this in the controller. Services should deal with the business logic, but I think this is more of an HTTP-layer concern.

dajinchu: Actually, leaving it in the service might make testing more feasible. I think using the `https` or `http` lib is fine - they're actually part of the node standard library so we don't need to add any packages.

packages/server/src/login/login.controller.ts
```diff
@@ -19,15 +21,23 @@ import {
...
 import { ConfigService } from '@nestjs/config';
 import { LoginCourseService } from './login-course.service';
 import { CourseSectionMappingModel } from './course-section-mapping.entity';
+import * as httpSignature from 'http-signature';
+import * as path from 'path';
+import * as fs from 'fs';
 
 @Controller()
 export class LoginController {
+  PUBLIC_KEY = fs.readFileSync(
```
liustanley, line 36: i think cypress is erroring on this line just fyi

liustanley (approved):
lgtm!

packages/server/src/login/login.controller.ts
willstenzel: thanks for the heads up!

PR #264, opened 2020-09-07
Allow TA's to choose preferred course section

### LINKED ISSUE #198: Allow TAs to choose course section
**Description:**
In our app, we can treat each section as a distinct "Course", but we need to find out which section a TA belongs to. For example, when a Fundies 1 TA logs in, they should be brought back to the `/` path where they can choose which course group they belong to (Regular, Accelerated, or NU Start). After selecting a course group, that course is set in their local storage, and navigating to our site defaults to that course from then on.

**Specs:**
- [ ] On TA login, they get redirected to `/` and we check local storage for existing selected course group
- [ ] If course group is present, simply redirect them to that course's id
- [ ] If no course group is present, display the course group selection UI
- [ ] Once they select a course, add it to local storage

If blocked, contact @liustanley

liustanley:
my code is probably really bad so let me know if anything needs to be changed

Closes #198


willstenzel:
Is this alright to close?

PR #269, opened 2020-09-08
Redirect From Login/NoCourses if a user has courses or is logged in

aislinblack:


willstenzel (approved):
Looks great! Should be good to go after @liustanley implements the redirect from that last saved course in `localstorage`

NEUDitao (approved):

packages/app/pages/login.tsx
```diff
@@ -13,6 +17,29 @@ const ContentContainer = styled.div`
...
+          defaultCourse !== null
+            ? defaultCourse.id
+            : profile.courses[0].course.id
+        }/today`
+      );
+      return "";
+    }
+    Router.push("nocourses");
+    return "";
+  }
```
dajinchu, line 41: Is it possible to generalize this and the one in `nocourses` into a hook?

dajinchu (changes requested):

dajinchu (approved):

PR #290, opened 2020-09-10
changed all avatars to initials + colors

NEUDitao:
they're beautiful I think

### DIFF
--- a/packages/app/components/Queue/TA/TABanner.tsx
+++ b/packages/app/components/Queue/TA/TABanner.tsx
@@ -1,4 +1,4 @@
-import { CheckOutlined, CloseOutlined, UserOutlined } from "@ant-design/icons";
+import { CheckOutlined, CloseOutlined } from "@ant-design/icons";
 import { API } from "@koh/api-client";
 import {
   ClosedQuestionStatus,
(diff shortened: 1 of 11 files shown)


willstenzel (approved):
Good stuff!

Only thing is we might want to change the font on the large avatars (like the ones on the today page) to have the same text to size ratio as the profile. @liustanley might have an opinion on that

image

liustanley (approved):
big brain looks good

i agree with will, just fix the text ratio on the today page and it's good to merge

liustanley (changes requested):
wait nvm lol some changes that need to be made:
- [ ] Make standard text to avatar size ratio, consistency here is important
- [ ] Missed the avatar on `TAQueueCard`

liustanley (approved):

packages/app/components/Queue/TA/TABanner.tsx
```diff
@@ -48,11 +50,20 @@ export default function TABanner({
...
-              icon={<UserOutlined />}
-              src={helpingQuestion.creator.photoURL}
-            />
+              style={{
+                backgroundColor: nameToRGB(helpingQuestion.creator.name),
+                fontSize: 36,
+              }}
+            >
+              {getInitialsFromName(helpingQuestion.creator.name)}
+            </Avatar>
```
dajinchu, line 66: Very similar Avatar code appears 7 times. Is it possible to abstract?

dajinchu (approved):

PR #296, opened 2020-09-13
TAs Notified when student joins empty queue

### LINKED ISSUE #295: Notify TA's when a student joins an empty queue

NEUDitao:
I told myself I'd start maintaining a healthy sleep schedule someday

Next thing I knew I was implementing this

closes #295

### DIFF
--- a/packages/server/src/question/question.controller.ts
+++ b/packages/server/src/question/question.controller.ts
@@ -132,24 +132,6 @@ export class QuestionController {
       isOnline: true,
     }).save();
 
-    if (
-      queue.questions.filter(
-        (question) =>
-          question.status !== OpenQuestionStatus.TADeleted &&
-          (Object.values(OpenQuestionStatus) as string[]).includes(
-            question.status,
-          ),
-      ).length === 0
-    ) {
-      // TODO: when limbo queue is implemented, don't need to do check for TA Deleted :)
-      queue.staffList.forEach((staff) => {
-        this.notifService.notifyUser(
-          staff.id,
-          NotifMsgs.ta.STUDENT_JOINED_EMPTY_QUEUE,
-        );
-      });
-    }
-
     return question;
   }
 
--- a/packages/server/src/question/question.controller.ts
+++ b/packages/server/src/question/question.controller.ts
@@ -132,6 +132,24 @@ export class QuestionController {
       isOnline: true,
     }).save();
 
+    if (
+      queue.questions.filter(
+        (question) =>
+          question.status !== OpenQuestionStatus.TADeleted &&
+          (Object.values(OpenQuestionStatus) as string[]).includes(
+            question.status,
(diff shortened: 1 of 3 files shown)


packages/server/src/question/question.controller.ts
```diff
@@ -132,6 +132,24 @@ export class QuestionController {
       isOnline: true,
     }).save();
 
+    if (
```
dajinchu, line 135: We should probably do this inside question subscriber, and use `QuestionModel.openInQueue` to get the list of open questions (faster and consistent with rest of app)

packages/server/src/question/question.subscriber.ts
```diff
@@ -62,6 +63,29 @@ export class QuestionSubscriber
...
+      })
+      .getCount();
+
+    const staff = (
+      await QueueModel.findOne(event.entity.queueId, {
+        relations: ['staffList'],
+      })
+    ).staffList;
+
+    if (numberOfQuestions === 0) {
```
liustanley, line 80: won't this never notify since this is run after insert? i'm not sure how this subscriber works, but i'm not getting the notification when testing

liustanley (changes requested):

packages/server/src/question/question.subscriber.ts
NEUDitao: DM'd, but in my testing (console logs) I was able to see that there were no questions here, and then I got a notif only on the first student.

NEUDitao: survey says it works on local but not vercel 👊

packages/server/src/question/question.subscriber.ts
```diff
@@ -62,6 +63,29 @@ export class QuestionSubscriber
...
   async afterInsert(event: InsertEvent<QuestionModel>): Promise<void> {
+    const numberOfQuestions = await QuestionModel.openInQueue(
+      event.entity.queueId,
+    )
+      .where('question.status IN (:...openStatus)', {
+        openStatus: [OpenQuestionStatus.Drafting, OpenQuestionStatus.Queued],
+      })
+      .getCount();
+
+    const staff = (
```
dajinchu, line 74: Can you grab the stafflist inside the if statement so we don't waste a db call when there's no need to notify

packages/server/src/question/question.subscriber.ts
dajinchu: It's because we're not running inside the same db transaction as the insertion, so we see the state of the db before insertion

dajinchu (changes requested):
Small bit and we're good

willstenzel (approved):
After @dajinchu's requested changes go through, looks like it's good to ship!

dajinchu (approved):

PR #308, opened 2020-09-15
tests validating hotfix 8a0e64dc0

dajinchu:
Ran tests on pre-hotfix code: 2 fails.
Ran tests on new code: success :)

### DIFF
--- a/packages/server/test/question.integration.ts
+++ b/packages/server/test/question.integration.ts
@@ -198,16 +198,64 @@ describe('Question Integration', () => {
         status: OpenQuestionStatus.Drafting,
       });
 
-      const response = await supertest({ userId: user.id })
+      await supertest({ userId: user.id })
         .post('/questions')
         .send({
           text: 'i need to know where the alamo is',
           queueId: queue.id,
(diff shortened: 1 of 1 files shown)


packages/server/test/question.integration.ts
```diff
@@ -198,16 +198,64 @@ describe('Question Integration', () => {
...
+        .post('/questions')
+        .send({
+          text: 'i need to know where the alamo is',
+          queueId: queue.id,
+          questionType: QuestionType.Bug,
+          force: false,
+        })
+        .expect(201);
+    });
+    it('works when other queues and courses exist', async () => {
```
willstenzel, line 238: Do we need this test if "other queue's and courses" are present in the previous test? It doesn't hurt, just seems like it might already be covered

willstenzel (approved):

NEUDitao (approved):

packages/server/test/question.integration.ts
dajinchu: yeah, when they failed, they failed differently (404 vs 401) so I made both tests

PR #309, opened 2020-09-15
Fix Prod Performance

dajinchu:
- Queue cleaning more efficient
- Queue cleaning done in background w/o blocking the TA delete endpoint
- No more eager loading ALL of a queue's questions

### DIFF
--- a/packages/server/src/course/course.controller.ts
+++ b/packages/server/src/course/course.controller.ts
@@ -119,6 +119,6 @@ export class CourseController {
     setTimeout(async () => {
       await this.queueCleanService.cleanQueue(queue.id);
       await this.queueSSEService.updateQueue(queue.id);
-    }, 1);
+    });
   }
 }
--- a/packages/server/src/course/course.controller.ts
+++ b/packages/server/src/course/course.controller.ts
@@ -49,9 +49,9 @@ export class CourseController {
 
     course.queues = await async.filter(
       course.queues,
-      async (q) => await q.checkIsOpen(),
+      async q => await q.checkIsOpen(),
     );
-    await async.each(course.queues, async (q) => await q.addQueueTimes());
+    await async.each(course.queues, async q => await q.addQueueTimes());
 
     return course;
   }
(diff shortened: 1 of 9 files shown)


packages/server/src/course/course.controller.ts
```diff
@@ -113,7 +116,9 @@ export class CourseController {
     }
     await queue.save();
     // Clean up queue if necessary
-    await this.queueCleanService.cleanQueue(queue.id);
-    await this.queueSSEService.updateQueue(queue.id);
+    setTimeout(async () => {
```
dajinchu, line 119: Use setTimeout so that we can return and send response to client, and do the cleaning later (1ms later)

packages/server/src/queue/queue.entity.ts
```diff
@@ -82,20 +79,10 @@ export class QueueModel extends BaseEntity {
...
-    }
-    return this.questions?.filter(
-      (q) =>
-        q.status === OpenQuestionStatus.Drafting ||
-        q.status === OpenQuestionStatus.Queued,
-    ).length;
+  queueSize: number;
+
+  async addQueueSize(): Promise<void> {
+    this.queueSize = await QuestionModel.openInQueue(this.id).getCount();
```
dajinchu, line 85: The way this is written, `queueSize` now includes all open questions, even `TADeleted` etc. I think we should just revamp all this in the upcoming Priority Queue ticket

packages/server/src/queue/queue.entity.ts
```diff
@@ -82,20 +79,10 @@ export class QueueModel extends BaseEntity {
     return open;
   }
 
-  @Expose()
-  get queueSize(): number {
```
dajinchu, line 86: 🚨 queueSize is no longer automatically sent to the client. **You must call `addQueueSize` if you want the queueSize present.**

liustanley (approved):
not able to test because of work but this looks great!

NEUDitao (approved):

willstenzel (approved):

PR #310, opened 2020-09-15
Fix queueSize to only count Drafting and Queued questions

willstenzel:


liustanley (approved):
lgtm

dajinchu (approved):

packages/server/src/queue/queue.entity.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+            status: OpenQuestionStatus.Drafting
+        });       
+        await QuestionFactory.create({
+            queue: queueFactory,
+            status: OpenQuestionStatus.Helping
+        });       
+        await QuestionFactory.create({
+            queue: queueFactory,
+            status: ClosedQuestionStatus.Resolved
+        });
```
dajinchu, line 45: super nitpick, but I think there's a `QuestionFactory.createList` that we could use

dajinchu (approved):

packages/server/src/queue/queue.entity.spec.ts
willstenzel: I don't think the `.createList` function allows us to create a list of questions each with different attributes. 

https://github.com/antonversal/typeorm-factory/blob/ebc7a1a1c70f8053dda7a5225d26d1a152715157/src/Factory.ts#L139-L147

PR #313, opened 2020-09-16
Push Built Files to Deployment Targets

dajinchu:
Deploying to staging/prod can take 5-10 minutes. Most of this is running `yarn build` which is super slow on the VMs and can even affect webapp performance. Solution is to build first, then push the built files to where we want to deploy them. 

_(we could do this with docker, but it's much simpler this way)_

PM2 pulls down from git, so we're just gonna host the built files **on a Git branch**. 

## Primary Changes
1. The `dist` branch has a different `.gitignore` that lets the built js files get committed.
2. The `deploy` GitHub Action runs `yarn build` and commit/pushes the result to the `dist` branch
3. Then, the action runs pm2 deploy the same way as before, except `ecosystem.config.js` gets `dist` instead of `master`, and doesn't run build in the `pre-deploy`.

## How does this affect the deployment process?

You can no longer run `yarn deploy production` immediately after pushing to master. Check GitHub to make sure the `Deploy > Build Dist Branch` job passed so that the changes will be on `dist`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,7 +3,7 @@ name: Deploy Staging
 on:
   push:
     branches:
-      - dc-build-on-gh
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


liustanley (approved):
this is galaxy brain omg

willstenzel (approved):
Big hype 🤯

PR #314, opened 2020-09-16
Disable Checkout Button when TA is helping a student

### LINKED ISSUE #286: Prevent "Check Out" while helping a student
When a TA is helping a student, they shouldn't be able to check out of the queue.

If blocked, contact @liustanley

willstenzel:


liustanley (approved):

packages/app/components/Queue/TA/TAQueueList.tsx
```diff
@@ -181,6 +181,7 @@ export default function TAQueueList({
                 {isStaffCheckedIn ? (
                   <CheckOutButton
                     danger
+                    disabled={isHelping}
```
dajinchu, line 184: Should we do a tooltip to explain? Normally we tooltip disabled buttons.
eg
https://github.com/sandboxnu/office-hours/blob/d3e7dd18abcf60b98863cd15c384846bd09a5d68/packages/app/components/Queue/TA/TAQueueList.tsx#L162-L166

dajinchu (changes requested):

packages/app/components/Queue/TA/TAQueueList.tsx
liustanley: rip lol

we were running into issues because of the styled component and decided it would be better in the interest of time to leave out the tooltip, not sure if there's a good way to implement it

willstenzel: Yeah unfortunately the tooltip was not playing nice with what we had. I think because the button was a nested `QueueInfoColumnButton`. 
```
const CheckOutButton = styled(QueueInfoColumnButton)`
  color: #da3236;
`;
```

When we moved the styles from the `QueueInfoColumnButton` to the `CheckOutButton` and made it a `styled(Button)` the tooltip worked but that seemed too hacky and would make the code worse.

dajinchu (approved):

packages/app/components/Queue/TA/TAQueueList.tsx
dajinchu: Seems right. We'll have to revisit this eventually

PR #342, opened 2020-09-21
Fix "cannot read property split of undefined"

liustanley:
I defaulted to the first color when the name string is undefined, but if there's a cleaner way to do this lmk


willstenzel (approved):
I'm assuming the name is undefined in a loading state so another option would be to default the background color to grey since there is already a lot of other grey on the page during loading. 

Either way works though.

dajinchu (approved):
go go go

PR #346, opened 2020-09-22
Check if name is defined before calling match

willstenzel:
I was seeing this error on Elastic APM so I thought this might fix it. 

image


NEUDitao (approved):

dajinchu (approved):
I think you might be right. The `split` error was also caused by a lack of name

PR #354, opened 2020-09-24
Better IsProd Checking

dajinchu:
**READ THIS PR CAREFULLY**

If a boolean is miffed somewhere in this PR we may open up dev features into prod :| 

The goal here is to avoid the isProd logic being spread out everywhere and get it right, in one spot. Also *actually* hide the dev panel correctly.

I'm starting to think it might be better to have an isProd on the frontend and one on the backend, cause reading process.env inside `common` is just a bit scary. Would like people's thoughts!

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -12,6 +12,9 @@ import { Type } from "class-transformer";
 import "reflect-metadata";
 
 export const PROD_URL = "https://khouryofficehours.com";
+export const isProd = (): boolean =>
+  process.env.DOMAIN === PROD_URL ||
+  (typeof window !== "undefined" && window?.location?.origin === PROD_URL);
 
 /////////////////////////
 // API Base Data Types //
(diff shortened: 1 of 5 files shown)


packages/common/index.ts
```diff
@@ -12,6 +12,9 @@ import { Type } from "class-transformer";
 import "reflect-metadata";
 
 export const PROD_URL = "https://khouryofficehours.com";
+export const isProd = (): boolean =>
+  process.env.DOMAIN === PROD_URL ||
+  (typeof window !== "undefined" && window?.location?.origin === PROD_URL);
```
willstenzel, line 17: I don't see anything wrong with this code, however my javascript knowledge is limited so I don't feel too qualified to speak on this. 

Even though it does seem a little sketchy to be checking `isProd` for both the client and server in one function, I like the simplicity of it.

NEUDitao (approved):

willstenzel (approved):

PR #361, opened 2020-09-30
Changed default question type from 'Bug' to 'Drafting' since it makes…

NEUDitao:
… more sense as a default value without having to hcange too much of the API

image


willstenzel (approved):
This totally works but I'm wondering if we would have better data consistency for the data analytics if we allowed the drafting question to have an empty `questionType`?

PR #362, opened 2020-09-30
Comment out location on popup card and remove unused file

willstenzel:
Just noticed we still show the location in the popup card so thought I'd remove it real quick.

image


liustanley (approved):
nice!

dajinchu (approved):

PR #364, opened 2020-10-02
Per-Device Notifs

dajinchu:
## Regular Usage
devices

## When User Does Not Allow Notif
device not allowed

## In Incognito
image

# Implementation

- Added code to send browser/OS info when registering for desktop notif.
- Added endpoint to delete your desktopnotif
- Return list of desktopnotifs in profile endpoint (redacted the sensistive info)
- Frontend just renders a list accordingly

### DIFF
--- a/packages/app/components/Nav/NotificationSettingsModal.tsx
+++ b/packages/app/components/Nav/NotificationSettingsModal.tsx
@@ -52,6 +52,7 @@ export function NotificationSettingsModal({
         const value = await form.validateFields();
         try {
           await editProfile(value);
+          form.setFieldsValue(profile);
           onClose();
         } catch (e) {
           if (
--- a/packages/app/components/Nav/NotificationSettingsModal.tsx
+++ b/packages/app/components/Nav/NotificationSettingsModal.tsx
@@ -130,6 +130,17 @@ function useThisDeviceEndpoint(): null | string {
   return endpoint;
 }
 
+function renderDeviceInfo(
+  device: DesktopNotifPartial,
+  isThisDevice: boolean
+): string {
+  if (device.name) {
+    return isThisDevice ? `${device.name} (This Device)` : device.name;
+  } else {
+    return isThisDevice ? "This Device" : "Other Device";
+  }
+}
+
 function DeviceNotifPanel() {
   const thisEndpoint = useThisDeviceEndpoint();
   const { data: profile, mutate } = useSWR(`api/v1/profile`, async () =>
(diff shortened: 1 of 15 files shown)


packages/app/components/Nav/NotificationSettingsModal.tsx
```diff
@@ -67,34 +80,17 @@ export function NotificationSettingsModal({
...
-                            ? "Please allow notifications in this browser"
-                            : "Browser does not support notifications. Please use Chrome or Firefox.",
-                        ],
-                      },
-                    ]);
-                  }
-                }
-              }}
-            />
+            <Switch />
```
willstenzel, line 88: Do we still need this? Since user can now add/remove individual devices it seems like it might confuse some people if they accidentally forgot to toggle the switch on.

packages/app/components/Nav/NotificationSettingsModal.tsx
```diff
@@ -125,3 +121,90 @@ export function NotificationSettingsModal({
...
+
+function renderDeviceInfo(
+  device: DesktopNotifPartial,
+  isThisDevice: boolean
+): string {
+  if (device.name) {
+    return isThisDevice ? `${device.name} (This Device)` : device.name;
+  } else {
+    return isThisDevice ? "This Device" : "Other Device";
+  }
```
willstenzel, line 141: Just curious when would we reach this `else` state? Is it for the legacy desktop notifications?

willstenzel (approved):
Bro why you got such a wrinkly brain lol

Seriously though, good stuff

willstenzel:
Also before this goes live we should add it to the notion [Help Guide](https://www.notion.so/Office-Hours-Help-Guide-a89c73dd53204cc3970ac44d61917417)

If people experience any weird issues it would be cool if we could add a link to the specific page in the help guide from the notifications modal and maybe even include a _troubleshooting_ section

NEUDitao (approved):

packages/app/components/Nav/NotificationSettingsModal.tsx
dajinchu: yep for legacy!

packages/app/components/Nav/NotificationSettingsModal.tsx
dajinchu: yeah I'll change it to default on, and run a sql query on prod to turn current user's on

packages/common/index.ts
```diff
@@ -29,17 +29,30 @@ export const isProd = (): boolean =>
...
+export class User {
+  id!: number;
+  email!: string;
+  name!: string;
+  photoURL!: string;
+  courses!: UserCourse[];
+  desktopNotifsEnabled!: boolean;
+
+  @Type(() => DesktopNotifPartial)
+  desktopNotifs!: DesktopNotifPartial[];
```
dajinchu, line 43: This is key: the profile endpoint now tells the frontend a list of the subscribe devices

packages/server/src/notification/desktop-notif.entity.ts
```diff
@@ -33,4 +32,10 @@ export class DesktopNotifModel extends BaseEntity {
 
   @Column({ nullable: true })
   userId: number;
+
+  @CreateDateColumn({ type: 'timestamp' })
+  createdAt: Date;
+
+  @Column({ type: 'text', nullable: true })
+  name: string;
```
dajinchu, line 40: Added a name field. When someone subscribes, the frontend will also send over a name like "Firefox on Ubuntu"

PR #371, opened 2020-10-07
Fix rendering for TA helping flow

willstenzel:
Before:
before_fix

After:
after

### DIFF
--- a/packages/app/components/Queue/TA/TAQueueList.tsx
+++ b/packages/app/components/Queue/TA/TAQueueList.tsx
@@ -131,19 +131,6 @@ export default function TAQueueList({
     question: Question,
     status: QuestionStatus
   ) => {
-    let newQuestions;
-    if (status in ClosedQuestionStatus) {
-      newQuestions = questions?.filter(q => q.id !== question.id);
-    } else if (status === OpenQuestionStatus.Helping) {
-      newQuestions = questions?.map((q) =>
-        q.id === question.id ? { ...q, status, taHelped: { id: user.id } } : q
-      );
-    } else {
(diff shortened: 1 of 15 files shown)


packages/app/components/Queue/TA/TAQueueList.tsx
```diff
@@ -130,13 +131,23 @@ export default function TAQueueList({
     question: Question,
     status: QuestionStatus
   ) => {
+    let newQuestions;
+    if (status in ClosedQuestionStatus) {
+      newQuestions = questions?.filter(q => q.id !== question.id);
+    } else if (status === OpenQuestionStatus.Helping) {
+      newQuestions = questions?.map((q) =>
+        q.id === question.id ? { ...q, status, taHelped: { id: user.id } } : q
```
willstenzel, line 139: Updating the `taHelped` attribute with an object that contains the id allows us to mutate the state so that we can correctly find the `helpingQuestion` (line 103).

dajinchu:
I'm worried about the race condition where two TAs help at the same time. The local mutations will make them both think they are helping, until SWR refetches and updates. What if we just didn't bother with local mutation at all? So we don't do anything until after https://github.com/sandboxnu/office-hours/blob/86d08fc20a204707d34a0d687365d0c72e0b1cba/packages/app/components/Queue/TA/TAQueueList.tsx#L147-L149 is awaited, then we just call `mutate()` to trigger a refresh. We can also add a nice error message saying "another TA is already helping" by looking at the result of `API.questions.update`

liustanley (approved):

dajinchu (approved):
nice

PR #381, opened 2020-10-14
Fix bug causing all TA's to be checked out

willstenzel:


packages/server/src/queue/queue-clean/queue-clean.service.ts
```diff
@@ -32,9 +32,6 @@ export class QueueCleanService {
       relations: ['staffList'],
     });
 
-    queue.staffList = [];
-    await queue.save(); // TODO: somehow keep track of this in the future and tell the TAs
```
NEUDitao, line 36: so this is removing the functionality of checking TA's out at midnight -- but we should hotfix this first. We can take care of doing the checking out TA's at midnight again later. Tests might fail after removing this

NEUDitao (approved):

dajinchu:
thanks guys

dajinchu (approved):

PR #395, opened 2020-10-23
fix default course sync

dajinchu:
Sometimes, if a student drops a class or logs in as a different user, they will end up getting redirected to a defaultCourse that they are not a member of and encounter a weird blank screen with just the footer and navbar. This fixes that, although we probably also want to make a 404 page for when people try to view class they aren't a member of.


willstenzel (approved):
Great catch

liustanley (approved):

PR #402, opened 2020-10-30
Proxy notion api request through backend

willstenzel:


packages/common/index.ts
```diff
@@ -446,3 +446,8 @@ export interface TwilioBody {
   From: string;
   ApiVersion: string;
 }
+
+export interface GetReleaseNotesResponse {
+  releaseNotes: unknown;
```
willstenzel, line 451: The reason this is an `unknown` type is because we didn't want to install the react-notion library on the server just so we could use the `BlockMapType` that it provides

liustanley (approved):

packages/app/components/Today/ReleaseNotes.tsx
```diff
@@ -1,45 +1,24 @@
...
-          ].value.properties.title = [];
-          setNotionReleaseNotes(result);
-        },
-        (error) => {
-          console.log("Error fetching release notes", error);
-        }
-      );
+    (async () => {
+      const data = await API.releaseNotes.get()
+      setNotionReleaseNotes(data.releaseNotes as BlockMapType)
```
dajinchu, line 19: wait why not `useSWR`

dajinchu (changes requested):

packages/server/src/seed/seed.controller.ts
```diff
@@ -168,7 +168,7 @@ export class SeedController {
     return 'Data successfully seeded';
   }
 
-  @Get('fillQueue')
+  @Get('fill_queue')
```
dajinchu, line 171: What is the reason for this?

packages/app/components/Today/ReleaseNotes.tsx
willstenzel: It's been a minute since I used it so I totally forgot 🤦 

I'll fix that up right away

packages/server/src/seed/seed.controller.ts
willstenzel: I figured I'd reformat to use underscore so we could follow the convention we were using before
https://github.com/sandboxnu/office-hours/blob/b9457f6c5bd052b1bebf9f20ad5ed9ef265e87ed/packages/server/src/course/course.controller.ts#L62

packages/app/components/Today/ReleaseNotes.tsx
willstenzel: Fixed.

dajinchu (approved):

PR #406, opened 2020-11-02
Fix Prettier On GitHub

dajinchu:
GitHub Action now runs prettier **and commits the changes**. See 6544153b59af1d25d05afc530d3fc97f789639c9 for example. This PR is big as a result because it prettifies everything that has fallen through the cracks, but in the future, nothing un-pretty should end up in master again, even if the local post-commit hook breaks on people's computers. 

**Main impact is our GitHub diffs should be smaller and without the styling fluff, making PR reviews a lot easier.**

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,12 +8,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: install node v12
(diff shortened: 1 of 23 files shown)


willstenzel (approved):
This is will keep the diffs squeaky clean 🧼 

Thanks Big Jon

NEUDitao (approved):

PR #412, opened 2020-11-08
Webpack 5

dajinchu:
Updated to Next.js 10, Nest 7.5, and enabled Webpack 5 for bundling frontend and backend.

Webpack 5 introduces file system caching so rebuilding the project should be faster each time. I am hoping this speeds up our deploy script that runs on master, but I'm not sure.

Please test this PR by running `yarn dev` and seeing if it feels any faster when you change the frontend or backend. It is likely only a marginal improvement, and only on some rebuilds when the cache can be used.


dajinchu:
As Percy shows, the CSS got messed up by this PR somehow.

liustanley:
@dajinchu is this still work in progress? or can we close this

PR #414, opened 2020-11-14
Update mutation to fix rendering bug

willstenzel:
Notice how when we save the question the queue list updates correctly but the student banner does not. 
Nov-14-2020 16-31-33


dajinchu (approved):
thanks! youve got sharp eyes 👀

NEUDitao (approved):

PR #430, opened 2020-11-22
TA Events Table / Check-In Check out Times monitored

NEUDitao:
```
EventsModel:

 - Time { Date }
 - EventType { String, but it's just a fucking Enumerator }
 - TACheckinInStatsModel => ManyToOne on USerCourseModel
 - Add optional columsn based on EventType and what's needed for future -- etc. events on questions get question model
 

Thought process here is that USerCourseModel is where we actually store if someone's a TA or not
so it's better to get the stuff from there per Course.
```


packages/common/index.ts
```diff
@@ -96,6 +96,14 @@ export enum Role {
   PROFESSOR = "professor",
 }
 
+/**
```
dajinchu, line 99: I think we can move this to the eventmodel.entity since it is not shared with the frontend

packages/server/src/course/course.controller.ts
```diff
@@ -91,6 +93,13 @@ export class CourseController {
     queue.staffList.push(user);
     await queue.save();
 
+    const UCM = await UserCourseModel.findOne({ where: { user, courseId } });
+    await EventModel.create({
+      time: new Date(),
+      eventType: EventType.TA_CHECKED_IN,
+      user: UCM,
```
dajinchu, line 100: what if we stored the userId and courseId directly in the eventmodel instead? UserCourseModels get destroyed and recreated on login so storing a reference to the ucm may get fucked when someone logs in again. It would also be less querying this way.

dajinchu (changes requested):

packages/server/src/course/course.controller.ts
NEUDitao: are you saying just as numbers and not as relations?

dajinchu: i just mean a relation to user and a relation to course, but not a relation to UserCourse

dajinchu (approved):
awesome :)

packages/server/src/profile/user-course.entity.ts
```diff
@@ -32,4 +34,8 @@ export class UserCourseModel extends BaseEntity {
 
   @Column({ type: 'enum', enum: Role, default: Role.STUDENT })
   role: Role;
+
+  @Exclude()
+  @OneToMany((type) => EventModel, (event) => event.user)
+  events: EventModel[];
```
willstenzel, line 40: Does this work given that the event model does not have a `UserCourse` foreign key?

NEUDitao: ah shoot I fogot to remove this on this side -- i'll get it when I get home

willstenzel (approved):

PR #433, opened 2020-11-23
priority queue service unit tests

dajinchu:


NEUDitao (approved):
i wonder if we can start working on getting Jest graphs of code coverage -- I know that's not the greatest metric but it's not the worst one

willstenzel (approved):
Writing test ain't all that fun, so I extra appreciate you doing this

PR #440, opened 2020-11-28
dc list detail

dajinchu:
This is a frontend-only PR. The detail pane gives us space to add a table with the student's question history and more, but those will require backend upgrades.

image

The Vercel preview link works.

## Responsive

| Full | Compact | Split View | Mobile |
|---|----|----|----|
| image | image | Peek 2020-11-30 12-09 | Peek 2020-11-30 12-10 |


willstenzel (approved):
Great stuff overall. 

Design wise, I like how the page is laid out but I think it would look nice for the `TAQueueListItems` to be cards with rounded edges (similar to the `TAStatusCard`) so everything has a consistent feel.

NEUDitao (approved):

PR #442, opened 2020-11-28
El names migration

NEUDitao:
Names -> First & last names

### DIFF
--- a/packages/server/src/backfill/separate-first-last-names.command.ts
+++ b/packages/server/src/backfill/separate-first-last-names.command.ts
@@ -13,24 +13,16 @@ export class BackfillSeparateFirstLastNames {
     const users = await UserModel.find();
     users.forEach((user) => {
       try {
-        return {
-          ...user,
-          name: '',
-          firstName: user.name.split[0],
-          lastName: user.name.split[1],
-        };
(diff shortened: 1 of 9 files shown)


dajinchu (approved):

packages/server/src/backfill/separate-first-last-names.command.ts
```diff
@@ -0,0 +1,28 @@
...
+  })
+  async fix(): Promise<void> {
+    const users = await UserModel.find();
+    users.forEach((user) => {
+      try {
+        user.firstName = user.name.split(' ')[0];
+        user.lastName = user.name.split(' ').slice(1).join(' ');
+      } catch (e) {
+        user.firstName = user.name;
+      }
```
willstenzel, line 20: What if we created a list of names that error which we can log after the script completes? I'm hoping their will at most be only a few that error, and if we have the list we can look them up fix it manually.

willstenzel (approved):

packages/server/src/backfill/separate-first-last-names.command.ts
NEUDitao: good idea -- adding it

PR #459, opened 2020-12-12
Wrap Queue Notes Today Page

dajinchu:


liustanley:
Could we also fix this issue in this PR? It's slightly involved with the queue notes, maybe more so the list detail container

image (8)
image (7)

liustanley:
Also this oof
Screen Shot 2020-12-13 at 3 38 39 PM

NEUDitao (approved):
i agree with stanley's questions but I think you'll be able to handle it

willstenzel (approved):

PR #460, opened 2020-12-12
Sentry APM Testing

### LINKED ISSUE #463: Move APM to Sentry

dajinchu:
Sentry is a bit easier to use I think. They recently announced support for monitoring postgres within node, and lacking that was my only complaint with them previously, so we will try them now﻿

CLoses #463


packages/server/src/course/course.controller.ts
```diff
@@ -44,6 +45,7 @@ export class CourseController {
   @Get(':id')
   @Roles(Role.PROFESSOR, Role.STUDENT, Role.TA)
   async get(@Param('id') id: number): Promise<GetCourseResponse> {
+    throw new InternalServerErrorException();
```
dajinchu, line 48: TODO REMOVE THIS. ITS JUST FOR TESTING

packages/app/components/common/ErrorBoundary.tsx
```diff
@@ -17,8 +18,10 @@ export class ErrorBoundary extends React.Component<{}, EBState> {
   }
 
   componentDidCatch(error, errorInfo): void {
+    console.log("CATCH");
```
willstenzel, line 21: Do we need this?

willstenzel (approved):

packages/app/components/common/ErrorBoundary.tsx
dajinchu: good CATCH, thanks

NEUDitao (approved):

PR #467, opened 2020-12-16
CACHE YA LATER! (caching the heatmap computation every week) 

dfarooq610:


packages/server/src/course/heatmap.service.ts
```diff
@@ -228,6 +242,6 @@ export class HeatmapService {
     autoExit: true,
   })
   async create(): Promise<void> {
-    const h = await this.getHeatmapFor(4);
+    const h = await this._getHeatmapFor(4);
```
willstenzel, line 245: This wasn't from this PR but why is `4` is hardcoded in the `heatmap:generate` command?

willstenzel (approved):
Nice work! 

I'd be curious as to what happens to the performance of the endpoint once this goes to prod.

packages/server/src/course/heatmap.service.ts
NEUDitao: ^^^^ it seems to be a dev thing but... still concerning

NEUDitao (approved):

packages/server/src/course/heatmap.service.ts
dfarooq610: I think this was a command thing, in a separate PR I'm gonna update this command

PR #468, opened 2020-12-16
logout endpoint and functionality added

### LINKED ISSUE #254: Add logout endpoint
We'll need to:
- [x] Create a `/lougout` endpoint that removes the auth cookie from local storage
- [x] Hook it up to the Logout button

NEUDitao:
tests written, let's get this shit in


packages/server/src/login/login.controller.ts
```diff
@@ -165,4 +165,18 @@ export class LoginController {
...
+    const isSecure = this.configService
+      .get<string>('DOMAIN')
+      .startsWith('https://');
+    res
+      .cookie('auth_token', '', {
+        httpOnly: true,
+        secure: isSecure,
+        expires: new Date(Date.now() - 86400000),
+      })
+      .redirect(302, '/');
```
dajinchu, line 180: would it make sense to send them to `/login` instead?

NEUDitao: wouldn't this do that by default? i'm just letting the app's default behaviour decide

packages/server/src/login/login.controller.ts
```diff
@@ -165,4 +165,18 @@ export class LoginController {
...
       .redirect(302, '/');
   }
+
+  @Get('/logout')
+  async logout(@Res() res: Response): Promise<void> {
+    const isSecure = this.configService
+      .get<string>('DOMAIN')
+      .startsWith('https://');
+    res
+      .cookie('auth_token', '', {
```
willstenzel, line 175: Would it make sense to use `res.clearCookie` here?

https://expressjs.com/en/4x/api.html#res.clearCookie

NEUDitao: yeah probably

willstenzel (approved):

dajinchu (approved):

PR #470, opened 2020-12-16
heatmap command positional argument

dfarooq610:
Removes the hardcoded 4 from the old command and now the courseId can be specified as a positional argument


willstenzel (approved):
👍

NEUDitao (approved):

PR #471, opened 2020-12-16
removed useless cypress fixtures

NEUDitao:
there were some old cypress fixtures from the first set of cypress tests that @willstenzel and I made before we knew how cypress like, actually worked. They didn't do anything but made the queue notes tests look out of place, so I've removed them and the `fixtures` folder that used to house them.


willstenzel (approved):
Looks good. I also removed the unused custom commands like the one used in `cy.mock`

dfarooq610 (approved):

PR #476, opened 2020-12-18
JWT Expiry

### LINKED ISSUE #425: JWT Expiry
We need to add an expiry to our JWTs so in the event that one gets leaked, it will become useless after a bit. This means making users log back in once in a while.

Plan:
- [x] Set a 1 month expiry to JWTs
- [x] Login endpoint: Don't override people's names when they log in again.
- [ ] Frontend: When user is not logged in, but they've logged in in the past, skip the /login page and go straight to Khoury Admin (saves people a click)
- [ ] After deploy: Change the JWT secret so that all existing JWTs become invalidated. (forces everyone to log in again and receive new JWT's with expiry)

Extension:
- [ ] For use during incidents, have a mechanism to revoke all of a user's JWT tokens (keep a table of revoked tokens). Connect it to the Admin Panel

dajinchu:
Closes #425 

Didn't implement thing where frontend goes straight to khoury. afraid it might be confusing. login page could def be more aesthetically pleasing tho since people will see it slightly more often


packages/server/src/login/login.controller.ts
```diff
@@ -169,7 +103,11 @@ export class LoginController {
 
   // Set cookie and redirect to proper page
   private async enter(res: Response, userId: number) {
-    const authToken = await this.jwtService.signAsync({ userId });
+    // Expires in a day
+    const authToken = await this.jwtService.signAsync({
+      userId,
+      expiresIn: 60 * 60 * 24 * 30,
+    });
```
willstenzel, line 110: Does this expire in a day or a month? A day seems pretty short

willstenzel (approved):
Thanks for moving the logic into the service. That definitely cleans things up

NEUDitao (approved):

packages/server/src/login/login.controller.ts
dajinchu: good catch. 30 days

PR #479, opened 2020-12-18
Added Professor Login

### LINKED ISSUE #474: Support instructor login
- [ ] When a professor or course coordinator logs in to the app, their user course for the course they teach should be set to professor

@willstenzel can you drop the log in schema returned by Khoury Admin here?

NEUDitao:
image

Closes #474

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -317,6 +317,9 @@ export class KhouryDataParams {
   @IsInt()
   campus!: string;
 
+  @IsInt()
+  professor!: string;
+
   @IsOptional()
   @IsString()
   photo_url!: string;
--- a/packages/server/src/seed/seed.controller.ts
+++ b/packages/server/src/seed/seed.controller.ts
@@ -146,6 +146,20 @@ export class SeedController {
         role: Role.TA,
         course: course,
       });
+      // Professor (Snarky!!)
+      const user5 = await UserFactory.create({
+        email: 'li.edwa@northeastern.edu',
+        name: 'Eddy Li',
+        firstName: 'Eddy',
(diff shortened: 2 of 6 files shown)


packages/server/src/seed/seed.controller.ts
```diff
@@ -146,6 +146,20 @@ export class SeedController {
...
       });
+      // Professor (Snarky!!)
+      const user5 = await UserFactory.create({
+        email: 'li.edwa@northeastern.edu',
+        name: 'Eddy Li',
+        firstName: 'Eddy',
+        lastName: 'Li',
+        photoURL:
+          'https://ca.slack-edge.com/TE565NU79-UR6P32JBT-a6c89822c544-512',
+      });
```
dfarooq610, line 157: DR LI IN THE HOUSE

dfarooq610 (approved):

packages/common/index.ts
```diff
@@ -317,6 +317,9 @@ export class KhouryDataParams {
   @IsInt()
   campus!: string;
 
+  @IsInt()
+  professor!: string;
+
```
willstenzel, line 322: I'm not exactly sure how validation works shouldn't the `IsInt` decorator match up with a `number` type?

willstenzel (changes requested):
Would you be able to add test to show that a new professor is created when the professor flag is present?

packages/common/index.ts
NEUDitao: I'm using line 318 as a precedent here -- it also uses `@IsInt()` on a string.

willstenzel (approved):

PR #480, opened 2020-12-19
Place In Line Improvement (?)

dajinchu:
Attempt at showing avatar with place-in-line at compact layouts. Previously the avatar would be hidden in narrow layouts, which was annoying to deal with. This gets rid of the "compact" breakpoint. I think it is simpler, and maybe better looking? _just worried it's confusing?_

| split list/detail | side-by-side list/detail |
| ---|---|
| image| image |


liustanley (approved):
This looks great! Overall I like that it simplifies our breakpoints, and saves a little bit of space. The badge animation is pretty cool too

NEUDitao (approved):

PR #481, opened 2020-12-19
Data Insights v0.1

willstenzel:


packages/app/components/Insights/components/SimpleChartComponent.tsx
```diff
@@ -0,0 +1,21 @@
+// import { Bar } from "@ant-design/charts";
+import { SimpleChartOutputType } from "@koh/common";
+import React, { ReactElement } from "react";
+
+interface SimpleChartTypes {
+  output: SimpleChartOutputType;
+}
+
+function SimpleChartComponent({ output }: SimpleChartTypes): ReactElement {
```
NEUDitao, line 9: what's the purpose of this? Is this something for the future?

packages/app/pages/course/[cid]/insights.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  if (!insight) {
+    // TODO: Create loading shell
+    return null;
+  }
+
+  // Determine which insight component to render
+  let InsightComponent;
+  switch (insight.component) {
+    case "SimpleDisplay":
```
Comment (NEUDitao) [inline packages/app/pages/course/[cid]/insights.tsx:53]: we should probably make SimpleDisplay/SimpleChart an enum...

this is one of the *very few* times PHP's "flexibility" would be helpful haha

packages/app/pages/course/[cid]/insights.tsx
```diff
@@ -0,0 +1,78 @@
...
+      InsightComponent = SimpleChartComponent;
+      break;
+    default:
+    // Return an error component
+    // Log error
+  }
+
+  return (
+    <Card
+      size="small"
```
Comment (NEUDitao) [inline packages/app/pages/course/[cid]/insights.tsx:66]: I'm worried that there are insights that aren't going to fit into this specific Insight size -- what's the plan in that case? For instance, things that require searching/scrolling through lists (TA Check In/Check Out is an example -- though that could also probably fit into this card size)

packages/app/pages/course/[cid]/insights.tsx
```diff
@@ -0,0 +1,78 @@
...
+  insightName: string;
+}
+
+function RenderInsight({ insightName }: RenderInsightProps): ReactElement {
+  const router = useRouter();
+  const { cid } = router.query;
+
+  const { data: insight, error, mutate } = useSWR(
+    cid && `api/v1/insights/${cid}/${insightName}`,
+    async () => API.insights.get(Number(cid), insightName)
```
Comment (NEUDitao) [inline packages/app/pages/course/[cid]/insights.tsx:42]: I'm *slightly* concerned about performance here -- since we're making a separate API request for each insight -- would it be better to batch them all up? @dajinchu would probably be the best person to ask (though, also, to be fair, this isn't a high-traffic endpoint, so we might be ok).

For once GraphQL would make sense here haha

packages/server/src/insights/insights.ts
```diff
@@ -0,0 +1,224 @@
...
+  Role,
+  PossibleOutputTypes,
+  SimpleDisplayOutputType,
+  SimpleChartOutputType,
+} from '@koh/common';
+import { UserCourseModel } from 'profile/user-course.entity';
+import { SelectQueryBuilder } from 'typeorm';
+import { QuestionModel } from 'question/question.entity';
+
+export interface InsightInterface<Model> {
```
NEUDitao, line 12: ya know I could fuck with OO JavaScript, this is pretty cool 😎

packages/server/src/insights/insights.ts
```diff
@@ -0,0 +1,224 @@
...
+      queryBuilder.where("role = 'student'"),
+      filters,
+    ).getCount();
+  }
+
+  addFilters(queryBuilder, filters): SelectQueryBuilder<UserCourseModel> {
+    filters.forEach((filter) => {
+      if (!this.possibleFilters.includes(filter.type)) {
+        throw new Error(
+          `${filter} is not a possbile filter for "${this.name}"`,
```
NEUDitao, line 54: why do we have to do this on every insight?

NEUDitao (commented):
Looks good! Just a few questions to ask -- I'll try to answer some time before my flight tn, but ping me on Slack if you need anything! I'll also take another re-look later when I get home or something

packages/app/pages/course/[cid]/insights.tsx
Comment (dajinchu) [reply packages/app/pages/course/[cid]/insights.tsx:42]: I think it's fine either way. Http overhead isn't that crazy. Having them separate might be better so they they can load in progressively, if one chart requires a ton of data it won't block the others from displaying

packages/app/components/Insights/components/SimpleChartComponent.tsx
willstenzel: The `SimpleChartComponent` we didn't quite get working because we were getting an error in the node module for @ant-design/charts which we were trying to use. We're still trying to nail down the best charting library to use. 

The reason we left this in here was because we wanted to show an example of how the output of an insight could render a chart.

packages/app/pages/course/[cid]/insights.tsx
Comment (willstenzel) [reply packages/app/pages/course/[cid]/insights.tsx:53]: Enum probably a good idea, but don't know a ton about how to use them. 

What would be the benefit of using an enum here?

packages/app/pages/course/[cid]/insights.tsx
Comment (willstenzel) [reply packages/app/pages/course/[cid]/insights.tsx:66]: My thinking was that we could set the size on the InsightComponent instead of the Card and which would allow the card to expand to whatever size the InsightComponent needs. 

Tbh I haven't fully thought through how all the styling is going to work. Hoping to talk with @liustanley about that soon.

packages/app/pages/course/[cid]/insights.tsx
Comment (willstenzel) [reply packages/app/pages/course/[cid]/insights.tsx:42]: We started with the endpoint retuning all the insights in one request but having the ability to load the page progressively we thought would be a better UX.

packages/server/src/insights/insights.ts
willstenzel: Good catch, I'll abstract that.

packages/app/pages/course/[cid]/insights.tsx
Comment (willstenzel) [reply packages/app/pages/course/[cid]/insights.tsx]: Added enum
https://github.com/sandboxnu/office-hours/pull/481/commits/21cdc0137f4e0c5ede1973d8ab94e6a7ea50275d

packages/server/src/insights/insights.ts
willstenzel: Filters abstracted.
https://github.com/sandboxnu/office-hours/pull/481/commits/360a511d6a6aefcf6d9ef365f84783200d5b6fce

dfarooq610: the display type is officially now an Enum!!

packages/app/pages/course/[cid]/insights.tsx
Comment (dfarooq610) [reply packages/app/pages/course/[cid]/insights.tsx:42]: Just for personal growth, why would GraphQL be beneficial here?

packages/app/components/Insights/components/SimpleChartComponent.tsx
NEUDitao: bet

packages/server/src/insights/insights.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+function addFilters(
+  queryBuilder,
+  modelName,
+  filters,
+  possibleFilters,
+): SelectQueryBuilder<QuestionModel> {
+  for (const filter of filters) {
+    if (!possibleFilters.includes(filter.type)) {
```
NEUDitao, line 35: what's preventing

```
for (const filter of filters) {
   if (possibleFilters.includes(filter.type)) {
       FILTER_MAP[modelName][filter.type](queryBuilder, filter);
   }
}
```

I feel like that'd be cleaner!

NEUDitao (approved):

packages/server/src/insights/insights.ts
willstenzel: You're totally right, just got that changed.

packages/app/package.json
```diff
@@ -10,18 +10,19 @@
     "tsc": "tsc"
   },
   "dependencies": {
+    "@ant-design/charts": "^1.0.11",
```
dajinchu, line 13: if this isn't working, maybe get rid of it from package.json

packages/server/src/insights/inisghts.controller.ts
```diff
@@ -0,0 +1,52 @@
+import { InsightsService } from './insights.service';
```
dajinchu, line 1: filename has typo

packages/server/src/insights/inisghts.controller.ts
```diff
@@ -0,0 +1,52 @@
...
+  ) {}
+
+  @Get(':courseId/:insightName')
+  @Roles(Role.PROFESSOR)
+  async get(
+    @Param('courseId') courseId: number,
+    @Param('insightName') insightName: string,
+    @Query() filters: any,
+  ): Promise<GetInsightResponse> {
+    // // TODO: Check that the current role is allowed to see the given insight
```
dajinchu, line 33: is this needed if the whole endpoint is guarded by role Professor?

packages/server/src/insights/insights.service.ts
```diff
@@ -0,0 +1,38 @@
...
+// interface generateAllInsightsParams {
+//   insights: InsightInterface<any>[];
+//   filters: any;  // TODO
+// }
+
+@Injectable()
+export class InsightsService {
+  constructor(private connection: Connection) {}
+
+  async generateOutput({ insight, filters }): Promise<any> {
```
dajinchu, line 14: some purpose statements on the functions in this service would help a lot with maintainability

dajinchu (approved):
nice work designing this system. Some small things to fix, but 👍🏼 for when you've done them.

packages/server/src/insights/inisghts.controller.ts
willstenzel: We definitely need one or the other. To make things extensible so that in the future we can allow certain insights to be shown to TA's as well, I'll remove the Professor role guard and change the code so the insight access up to the `roles` attribute of the insight itself.

NEUDitao:
One more comment -- would it be possible to put the insight names into an enum themselves or just constants in `common/index.ts`?

packages/server/src/insights/inisghts.controller.ts
willstenzel: oof.. good catch

packages/app/pages/course/[cid]/insights.tsx
Comment (NEUDitao) [reply packages/app/pages/course/[cid]/insights.tsx:66]: I think that'd be a great idea -- I'm down to add those to a PR after thsi one is merged and see what people think

packages/app/package.json
willstenzel: yeah unfortunately we could not get it working so I removed it. 

I still think there's hope, but we can explore again in the future

packages/server/src/insights/insights.service.ts
willstenzel: Good idea.

willstenzel:
> One more comment -- would it be possible to put the insight names into an enum themselves or just constants in `common/index.ts`?

@NEUDitao I'm not sure what exactly your looking for. I could add something like this but it would make accessing the insights more wordy. 
```
// common/index.js
export enum INSIGHT_NAMES {
  totalStudents,
  totalQuestionsAsked,
  averageWaitTime,
  questionTypeBreakdown,
}

// insights.ts
export const INSIGHTS_MAP = {
  [INSIGHT_NAMES.totalStudents]: new TotalStudents(),
  [INSIGHT_NAMES.totalQuestionsAsked]: new TotalQuestionsAsked(),
  [INSIGHT_NAMES.averageWaitTime]: new AverageWaitTime(),
  [INSIGHT_NAMES.questionTypeBreakdown]: new QuestionTypeBreakdown(),
};
```

NEUDitao:
> > One more comment -- would it be possible to put the insight names into an enum themselves or just constants in `common/index.ts`?
> 
> @NEUDitao I'm not sure what exactly your looking for. I could add something like this but it would make accessing the insights more wordy.
> 
> ```
> // common/index.js
> export enum INSIGHT_NAMES {
>   totalStudents,
>   totalQuestionsAsked,
>   averageWaitTime,
>   questionTypeBreakdown,
> }
> 
> // insights.ts
> export const INSIGHTS_MAP = {
>   [INSIGHT_NAMES.totalStudents]: new TotalStudents(),
>   [INSIGHT_NAMES.totalQuestionsAsked]: new TotalQuestionsAsked(),
>   [INSIGHT_NAMES.averageWaitTime]: new AverageWaitTime(),
>   [INSIGHT_NAMES.questionTypeBreakdown]: new QuestionTypeBreakdown(),
> };
> ```

Just worried since it's used there as well as the names of insights.ts, it's a lot of places to keep track of magic strings

PR #484, opened 2020-12-20
manually refresh office hours

### LINKED ISSUE #483: As a course coordinator, I want to manually refresh the office hours calendar
- [ ] New button for instructors to click which triggers a scrape for their class
- [ ] Maybe add a cooldown to prevent spamming this button?

NEUDitao:
image

Closes #483

### DIFF
--- a/packages/app/components/Schedule/SchedulePanel.tsx
+++ b/packages/app/components/Schedule/SchedulePanel.tsx
@@ -1,3 +1,4 @@
+import { Role } from "@koh/common";
 import moment from "moment";
 import React, { ReactElement } from "react";
 import {
@@ -9,6 +10,7 @@ import {
 } from "react-big-calendar";
 import styled from "styled-components";
 import { useCourse } from "../../hooks/useCourse";
+import { useRoleInCourse } from "../../hooks/useRoleInCourse";
 import UpdateCalendarButton from "./UpdateCalendarButton";
 
 const ScheduleCalendar = styled(Calendar)<CalendarProps>`
@@ -25,6 +27,7 @@ export default function SchedulePanel({
   defaultView = "week",
 }: ScheduleProps): ReactElement {
   const { course } = useCourse(courseId);
+  const role = useRoleInCourse(courseId);
 
   const myEvents: Event[] =
     course?.officeHours.map((e) => ({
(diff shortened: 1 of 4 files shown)


packages/app/components/Schedule/SchedulePanel.tsx
```diff
@@ -32,6 +52,29 @@ export default function SchedulePanel({
       title: e.title,
     })) ?? [];
 
+  const renderButton = () => {
```
dajinchu, line 55: Could we move this whole button to its own component? It adds 3 hooks to this component so it can clutter the rest of the schedulepanel

dajinchu (changes requested):

packages/app/components/Schedule/SchedulePanel.tsx
NEUDitao: done

packages/app/components/Schedule/SchedulePanel.tsx
```diff
@@ -43,6 +44,7 @@ export default function SchedulePanel({
           new Date(today.getFullYear(), today.getMonth(), today.getDate(), 8)
         }
       />
+      <UpdateCalendarButton courseId={courseId} />
```
willstenzel, line 47: Should we conditionally render based on `role === Role.PROFESSOR` here instead of rendering `null` in the UpdateCalendarButton component?

packages/app/components/Schedule/UpdateCalendarButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+export default function UpdateCalendarButton({
+  courseId,
+}: UpdateCalendarButtonProps): ReactElement {
+  const [updated, setUpdated] = useState(CalendarUpdateStatus.BEFORE);
+  const role = useRoleInCourse(courseId);
+
+  const updateCalendar = async () => {
+    setUpdated(CalendarUpdateStatus.UPDATING);
+    await API.course.updateCalendar(courseId);
+    setUpdated(CalendarUpdateStatus.AFTER);
```
willstenzel, line 26: Will the calendar actually update after this happens? Or do we need to refetch the `officeHours` for a course to make that happen?

NEUDitao: it updated the calendar when I updated on local :o

packages/app/components/Schedule/SchedulePanel.tsx
NEUDitao: yeah I was debating between having the hook in Schedule or UPdate, but I think it's also more readable if I put it in Schedule

willstenzel (approved):

dajinchu (approved):
epic 😎

PR #492, opened 2020-12-22
Zero Downtime

dajinchu:
# Goal
Use PM2 to run 2 instances of the app in cluster mode, load balancing between them. `prod:start` then uses `reload` instead of `restart`, which will restart them one at a time, waiting for one to come up before starting the next. This results in having 1 instance of the app up at all times.

We will now run 2 instances of the app in parallel. This will help us scale the app out horizontally in the future, to handle more users.

# Implications
HOWEVER, running a system like this means we can NO LONGER persist things in memory and expect them to be the same on subsequent requests, since requests are split across multiple instances that may restart randomly. This new constraint so far just affects SSE, so much of this PR is just rerwriting SSE to use Redis to pub/sub between instances of the backend and send SSE messages no matter which backend instance the clients are connected to.

# Deployment
Before deploying, we must install Redis on staging and production.
Upon deploying, we must do `yarn prod:stop` and `yarn prod:start` to fully refresh pm2.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,8 +24,6 @@ jobs:
       - run: echo "::set-output name=version::$(git rev-parse origin/master)"
         id: pre-version
       - uses: bahmutov/npm-install@v1
-      - name: Delete old builds
-        run: rm -rf packages/app/.next
       - run: yarn build
         env:
           NEXT_PUBLIC_APM_SERVER: ${{ secrets.ELASTIC_APM_SERVER_URL}}
--- a/infrastructure/prod/ecosystem.config.js
+++ b/infrastructure/prod/ecosystem.config.js
@@ -2,9 +2,11 @@ module.exports = {
   apps: [
     {
       name: "app",
-      script: "npm",
-      args: "run prod:start",
+      script: "./node_modules/.bin/next",
+      args: "start -p 3001",
       cwd: "./packages/app",
+      instances: 2,
+      exec_mode: "cluster",
       env: {
         NODE_ENV: "production",
       },
@@ -13,7 +15,7 @@ module.exports = {
       name: "server",
       cwd: "./packages/server",
       script: "dist/main.js",
-      node_args: "--max-old-space-size=8192",
+      node_args: "--max-old-space-size=4096",
       instances: 2,
       exec_mode: "cluster",
       env: {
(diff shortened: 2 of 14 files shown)


.github/workflows/deploy.yml
```diff
@@ -24,8 +24,6 @@ jobs:
       - run: echo "::set-output name=version::$(git rev-parse origin/master)"
         id: pre-version
       - uses: bahmutov/npm-install@v1
-      - name: Delete old builds
```
dajinchu, line 27: don't delete old builds, so that when deploying, we can continue running the old frontend in parallel with the new one

packages/app/pages/course/[cid]/today.tsx
```diff
@@ -68,8 +68,9 @@ export default function Today(): ReactElement {
     mutateCourse();
   };
 
+  console.log(course, profile);
   const queueCheckedIn = course?.queues.find((queue) =>
-    queue.staffList.find((staff) => staff.id === profile.id)
+    queue.staffList.find((staff) => staff.id === profile?.id)
```
Comment (dajinchu) [inline packages/app/pages/course/[cid]/today.tsx:72]: race condition exacerbated by the zero downtime deployment, since requests are more likely to arrive out of order

infrastructure/prod/ecosystem.config.js
```diff
@@ -2,18 +2,22 @@ module.exports = {
   apps: [
     {
       name: "app",
-      script: "npm",
-      args: "run prod:start",
+      script: "./node_modules/.bin/next",
+      args: "start -p 3001",
       cwd: "./packages/app",
+      instances: 2,
+      exec_mode: "cluster",
```
dajinchu, line 9: This is the core of it. We run 2 instances in cluster mode, which will load balance requests between the two.

packages/server/src/sse/sse.service.ts
```diff
@@ -1,34 +1,97 @@
...
  * T is metadata associated with each Client
  *
  * Low level abstraction for sending SSE to "rooms" of clients.
  * Probably don't use this directly, and wrap it in a service specific to that event source
+ *
+ * This handles when there's multiple backend instances by assigning unique client ids to each connection.
+ * When one instance wants to send to a client, it publishes to a Redis channel for the client.
+ * All instances listen to Redis, and if they are the one managing that client, they send the msg.
+ *
+ * Rooms with client metadata are also maintained in Redis key/value store.
```
dajinchu, line 32: This service got a good bit more complex, but read this comment to get the gist of it.

willstenzel (approved):
Some heat right here 🔥

NEUDitao (approved):

PR #494, opened 2020-12-22
Profile Pictures Added

NEUDitao:


packages/app/components/Settings/SettingsPage.tsx
```diff
@@ -27,23 +43,72 @@ const VerticalDivider = styled.div`
...
-            fontSize={avatarSize * (3 / 7)}
-          />
+          <>
+            {uploading && <Spin />}
+            {!uploading &&
+              (profile?.photoURL && error === undefined ? (
+                <Avatar
+                  icon={<UserOutlined />}
+                  src={"/api/v1/profile/get_picture/" + profile.photoURL}
+                  size={avatarSize}
```
NEUDitao, line 84: I plan to abstract this Avatar (and eventually all Avatars) to have the switching model

packages/server/src/profile/profile.controller.ts
```diff
@@ -3,13 +3,27 @@ import {
...
+  UseInterceptors,
+} from '@nestjs/common';
+import { FileInterceptor } from '@nestjs/platform-express';
 import { pick } from 'lodash';
 import { Connection } from 'typeorm';
 import { JwtAuthGuard } from '../login/jwt-auth.guard';
 import { NotificationService } from '../notification/notification.service';
 import { User } from './user.decorator';
 import { UserModel } from './user.entity';
+const fs = require('fs');
```
NEUDitao, line 26: doing import fs from 'fs', or import { fs } from 'fs'; causes and error and I don't quite know why but I'll figure that out when I have more sleep

packages/server/src/profile/profile.controller.ts
```diff
@@ -81,4 +95,39 @@ export class ProfileController {
 
     return this.get(user);
   }
+
+  @Post('/upload_picture')
+  @UseInterceptors(
+    FileInterceptor('file', {
```
dajinchu, line 101: Add `limits` https://github.com/expressjs/multer#multeropts to prevent us from getting blasted by large uploads. Frontend validation is good, but backend needs to be protected

packages/server/src/profile/profile.controller.ts
```diff
@@ -81,4 +95,39 @@ export class ProfileController {
...
+      dest: 'uploads/',
```
dajinchu, line 102: For compression, we probably don't want to save to file system, get the file, compress, and save to file system again. 

Maybe we switch to the Memory engine and do file compression + saving ourselves.
https://github.com/expressjs/multer#memorystorage

packages/server/src/profile/profile.controller.ts
```diff
@@ -81,4 +95,39 @@ export class ProfileController {
...
+    }),
+  )
+  async uploadImage(
+    @UploadedFile() file,
+    @User() user: UserModel,
+  ): Promise<void> {
+    if (user.photoURL) {
+      fs.unlinkSync('uploads/' + user.photoURL);
```
dajinchu, line 110: `sync` filesystem functions are slow af. pls await the async versions so our poor CPU can do other stuff while the hard drive seeks

packages/server/src/profile/profile.controller.ts
dajinchu: Also, use an envionment variable for the file upload location.

dajinchu (commented):
Great start

packages/app/components/Settings/SettingsPage.tsx
```diff
@@ -27,23 +43,72 @@ const VerticalDivider = styled.div`
...
       <Col span={4} style={{ textAlign: "center" }}>
         {avatarSize ? (
-          <AvatarWithInitals
-            style={{ marginTop: "60px", marginBottom: "60px" }}
-            name={profile?.name}
-            size={avatarSize}
-            fontSize={avatarSize * (3 / 7)}
-          />
+          <>
+            {uploading && <Spin />}
```
willstenzel, line 78: This might already be in your TODO's but I noticed that the spin icon renders way up at the top of the sidebar which causes all the buttons to move up right under it. It would be nice if the spinner could show were the avatar was previously. 

image

willstenzel (commented):
Overall looks solid - just had a few minor comments.

packages/server/src/profile/profile.controller.ts
NEUDitao: funnily enough https://github.com/nestjs/nest/blob/f472852d441b7f66e52449e09f885ab57807bf96/packages/platform-express/multer/interfaces/multer-options.interface.ts#L15 the default is 1 MB :o. would you still want it for clarity?

willstenzel (approved):
Awesome stuff!

packages/server/src/profile/profile.controller.ts
```diff
@@ -99,35 +105,57 @@ export class ProfileController {
...
     }),
   )
   async uploadImage(
-    @UploadedFile() file,
+    @UploadedFile() file: Express.Multer.File,
     @User() user: UserModel,
   ): Promise<void> {
     if (user.photoURL) {
-      fs.unlinkSync('uploads/' + user.photoURL);
+      fs.unlink(process.env.UPLOAD_LOCATION + '/' + user.photoURL, (err) => {
```
dajinchu, line 116: use `path.join` instead of concatenating `/` for windows support

packages/server/src/profile/profile.controller.ts
```diff
@@ -99,35 +105,57 @@ export class ProfileController {
...
   @Get('/get_picture/:photoURL')
   async getImage(
     @Param('photoURL') photoURL: string,
     @User() user: UserModel,
-    @Res() res,
+    @Res() res: Response,
   ): Promise<void> {
-    if (fs.existsSync('uploads/' + user.photoURL)) {
-      res.sendFile(user.photoURL, { root: 'uploads' });
+    if (fs.existsSync(process.env.UPLOAD_LOCATION + '/' + photoURL)) {
```
dajinchu, line 155: use async version

packages/server/src/profile/profile.controller.ts
```diff
@@ -99,35 +105,57 @@ export class ProfileController {
...
+          user.photoURL,
+          err,
+          'the previous image was at an invalid location?',
+        );
+      });
     }
 
-    user.photoURL = file.filename;
-    await user.save();
+    const spaceLeft = await checkDiskSpace('/');
```
dajinchu, line 126: root dir is diff on windows :( 
use this to detect root dir:
https://stackoverflow.com/a/50299531

packages/server/src/profile/profile.controller.ts
```diff
@@ -99,35 +105,57 @@ export class ProfileController {
...
-    console.log(test);
+    const fileName =
+      user.id +
+      '-' +
+      Math.random().toString(36).substring(2, 15) +
+      Math.random().toString(36).substring(2, 15);
+
+    await sharp(file.buffer)
+      .resize(256)
+      .toFile(process.env.UPLOAD_LOCATION + '/' + fileName);
```
dajinchu, line 143: path.join

dajinchu (approved):
Are we planning on downloading pics we get from khoury?

Just some small fixes. Looks good

packages/server/src/profile/profile.controller.ts
NEUDitao: does this matter since we're on the Khoury VMs?

dajinchu (approved):

packages/server/src/profile/profile.controller.ts
dajinchu: no but in case of windows developers

dajinchu:
Just update the jest snapshots I think

willstenzel (approved):

packages/server/src/profile/profile.controller.ts
NEUDitao: kill all windows developers

PR #498, opened 2020-12-25
fixed tests

NEUDitao:
Looksl ike names were missing at some point: o


packages/server/src/insights/insights.ts
```diff
@@ -22,6 +22,7 @@ export interface InsightInterface<Model> {
     insightFilters: any,
   ) => Promise<PossibleOutputTypes>;
   output?: Promise<PossibleOutputTypes>;
+  name: string;
```
NEUDitao, line 25: this is the meat of the PR

PR #505, opened 2020-12-29
Professor Office Hours

### LINKED ISSUE #501: Enable professor office hours
- [x] Update scraper to look for professor office hours with the keyword `Professor`
- [x] Professor office hours use a separate queue from TAs
- [x] Professor office hour queue location is the calendar event title (write tests for multiple professor queues)
- [x] TA check in button on Today page only checks TAs into TA queues
- [x] Professor check in button on Today page only checks professors into professor queues

Bonus:
- [ ] Professor check in button brings up the old modal that allows them to either join TA queue or create new professor queue

NEUDitao:
Closes #501


cypress/integration/common/priority_queue.spec.js
```diff
@@ -120,7 +120,7 @@ describe("TA Priority Queue", () => {
   });
 
   describe("Requeue Student", () => {
-    it.only("Requeue student works", function () {
+    it("Requeue student works", function () {
```
willstenzel, line 123: nice find! I wonder if there is a way to have the linter catch this in the future?

packages/server/src/course/ical.service.ts
```diff
@@ -155,9 +157,13 @@ export class IcalService {
       }).save();
     }
 
+    const officeHoursEventRegex = /\b^(OH|Hours)\b/;
```
willstenzel, line 160: Is there a reason to make this regex the same as the `testRegex` from above?

packages/server/src/course/course.controller.ts
```diff
@@ -59,10 +63,20 @@ export class CourseController {
...
+        courseId: id,
+      },
+    });
+
+    if (userCourseModel.role !== Role.PROFESSOR) {
+      course.queues = await async.filter(
+        course.queues,
+        async (q) => await q.checkIsOpen(),
+      );
+    }
```
willstenzel, line 78: Would it make sense to have the filter allow `isProfessorQueue` queues if the user is a professor instead of ignoring the filter all together? 

That way we can avoid having a large number of queues showing up for a professor when we switch back to in-person office hours.

packages/app/components/Today/ProfessorCheckinButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+        >
+          Check In
+        </CheckinButton>
+      )}
+      {queueCheckedIn && (
+        <TACheckinButton
+          courseId={Number(cid)}
+          room={queueCheckedIn.room}
+          state="CheckedIn"
+        />
```
willstenzel, line 71: Not a big deal but it could be nice to have a single `CheckinButton` that get's used in `today.tsx` which handles all the check-in/check-out logic for TA's and Professors

packages/server/src/course/ical.service.ts
NEUDitao: Oh we definitely don't need this now that `testRegex` has a default eh

packages/server/src/course/course.controller.ts
NEUDitao: that makes a lot more sense...

packages/app/components/Today/ProfessorCheckinButton.tsx
NEUDitao: I think I could manage that

cypress/integration/common/priority_queue.spec.js
NEUDitao: same

willstenzel (approved):

dajinchu (approved):
very nice

packages/app/components/Today/ProfessorCheckinButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+const ProfessorModalRadio = styled(Radio)`
+  display: block;
+  height: 30px;
+  lineheight: 30px;
+`;
+
+export default function TodayPageCheckinButton(): ReactElement {
+  const profile = useProfile();
+  const [modalVisible, setModalVisible] = useState(false);
+  const [value, setValue] = useState(0);
```
dajinchu, line 21: confusing variable name

PR #510, opened 2020-12-31
Adding timezones to courses

dfarooq610:
The only thing that might be a little weird with course timezones would be queue-cleans at midnight, but we should have a productive discussion about that because courses will now have office hours throughout the day across various countries. Closes #510

### DIFF
--- a/packages/server/src/backfill/backfill-course-timezones.ts
+++ b/packages/server/src/backfill/backfill-course-timezones.ts
@@ -3,7 +3,7 @@ import { Injectable } from '@nestjs/common';
 import { CourseModel } from 'course/course.entity';
 
 @Injectable()
-export class BackfillQuestionFirstHelpedAt {
+export class BackfillCourseTimezones {
   @Command({
     command: 'backfill:course-timezones',
     describe: 'set all course timezones from null to "America/New_York"',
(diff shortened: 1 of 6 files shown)


packages/server/src/backfill/backfill-course-timezones.ts
```diff
@@ -0,0 +1,20 @@
+import { Command } from 'nestjs-command';
+import { Injectable } from '@nestjs/common';
+import { CourseModel } from 'course/course.entity';
+
+@Injectable()
+export class BackfillQuestionFirstHelpedAt {
```
willstenzel, line 6: Rename this BackfillCourseTimezones?

willstenzel (approved):
Left a comment on one minor thing, but changes look good.

NEUDitao (approved):

packages/server/test/__snapshots__/course.integration.ts.snap
```diff
@@ -15,6 +15,7 @@ Object {
     },
   ],
   "queues": Array [],
+  "timezone": null,
```
NEUDitao, line 18: should we add a tz to the test? i know they're nullable, but in the future, our intention is for them to *not* be right?

NEUDitao (approved):

PR #519, opened 2021-01-09
Require TA to be checked in to remove student

willstenzel:


NEUDitao (approved):

packages/app/components/Queue/TA/TAQueueDetailButtons.tsx
```diff
@@ -117,18 +118,27 @@ export default function TAQueueDetailButtons({
...
-            />
+          <Tooltip
+            title={
+              isCheckedIn
+                ? "Remove From Queue"
+                : "You must check in to remove students from the queue"
+            }
+          >
+            <span>
+              {/* This span is a workaround for tooltip-on-disabled-button
```
dfarooq610, line 129: Does this tooltip work properly? @NEUDitao and I were messing around with tooltips on disabled buttons by wrapping them in a span for adhoc queues and it was just acting up way too much

dfarooq610 (approved):

packages/app/components/Queue/TA/TAQueueDetailButtons.tsx
willstenzel: It seemed to be working for me

PR #521, opened 2021-01-12
deploy script

dajinchu:


willstenzel (approved):

dajinchu:
> this is great! how should we handle backfills when deploying though?

Just gonna have to ssh and run them. Usually we should write features in a way that not running the backfill just results in degraded service

infrastructure/prod/deploy.sh
```diff
@@ -0,0 +1,11 @@
...
+# Runs deployment on this machine. Do NOT run this locally.
+set -ex
+
+cd /var/www/source
+git fetch
+git reset --hard origin/dist
+yarn
+yarn typeorm migration:run
+env HOME=/var/www pm2 startOrReload infrastructure/prod/ecosystem.config.js
+chgrp -R officehours /var/www/source/.pm2
```
liustanley, line 11: Oh wait this has to be `/var/www/.pm2`

liustanley (changes requested):

PR #522, opened 2021-01-13
Update Insights Style

willstenzel:
This PR makes it so that all the `small` insights will be rendered on top of the larger ones.  

image

### DIFF
--- a/packages/app/pages/course/[cid]/insights.tsx
+++ b/packages/app/pages/course/[cid]/insights.tsx
@@ -15,20 +15,34 @@ import InsightsDisplayOptions from "../../../components/Insights/components/Insi
 export default function Insights(): ReactElement {
   const router = useRouter();
   const { cid } = router.query;
-  const { data: profile, error, mutate } = useSWR(`api/v1/profile`, async () =>
-    API.profile.index()
+  const { data: profile, mutate: mutateProfile } = useSWR(
+    `api/v1/profile`,
+    async () => API.profile.index()
+  );
+  const { data: allInsights } = useSWR(`api/v1/insights/listAll`, async () =>
+    API.insights.list()
   );
   const [settingsVisible, setSettingsVisible] = useState(false);
 
   const toggleInsightOn = async (insightName) => {
     await API.insights.toggleOn(insightName);
-    mutate();
+    mutateProfile();
   };
 
   const toggleInsightOff = async (insightName) => {
     await API.insights.toggleOff(insightName);
-    mutate();
+    mutateProfile();
   };
+
+  if (!allInsights || !profile?.insights) {
+    return null;
+  }
+  // Split users insights into group by size so they can be rendered correctly
+  const [smallInsights, defaultInsights] = profile.insights.reduce(
+    ([s, d], i) =>
+      allInsights[i].size === "small" ? [[...s, i], d] : [s, [...d, i]],
+    [[], []]
+  );
 
   return (
     <>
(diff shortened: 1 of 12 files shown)


packages/app/pages/course/[cid]/insights.tsx
```diff
@@ -110,10 +135,15 @@ function RenderInsight({
     <Card
       size={insight.size as CardSize}
       title={insight.displayName}
+      style={{
+        margin: "8px",
+        width: insight.size === "default" ? "600px" : "200px",
+        height: insight.size === "default" ? "600px" : "150px",
```
Comment (willstenzel) [inline packages/app/pages/course/[cid]/insights.tsx:141]: Instead of setting pixel amounts would it be better to have the insights each take up a percentage of the screen?

Comment (NEUDitao) [reply packages/app/pages/course/[cid]/insights.tsx:141]: I think that'd be more sensible yeah

NEUDitao (approved):

packages/app/pages/course/[cid]/insights.tsx
```diff
@@ -15,21 +15,35 @@ import InsightsDisplayOptions from "../../../components/Insights/components/Insi
...
-    mutate();
+    mutateProfile();
   };
 
+  if (!allInsights || !profile?.insights) {
+    return null;
+  }
+  // Group users insights by size (small | default) so they can be rendered correctly
+  const [smallInsights, defaultInsights] = profile.insights.reduce(
+    ([s, d], i) =>
```
Comment (dfarooq610) [inline packages/app/pages/course/[cid]/insights.tsx:42]: Just for self actualization, What do s, d, and I represent?

packages/server/src/insights/insights.controller.ts
```diff
@@ -72,35 +72,28 @@ export class InsightsController {
...
     @Body() body: { insightName: string },
     @User() user: UserModel,
-  ): Promise<ListInsightsResponse> {
-    const updatedInsights = await this.insightsService.toggleInsightOff(
-      user,
-      body.insightName,
-    );
-    return updatedInsights;
+  ): Promise<void> {
+    await this.insightsService.toggleInsightOff(user, body.insightName);
```
dfarooq610, line 96: This is a JS question but since toggleInsightOff returns Promise<void> I think you can just say "return await this.insightsService..."

packages/server/src/insights/insight-classes.ts
```diff
@@ -166,11 +162,10 @@ export class AverageWaitTime implements InsightInterface<QuestionModel> {
   displayName = 'Average Wait Time';
   description = 'Gets the average wait time';
   roles = [Role.PROFESSOR];
-  component = InsightDisplay.SimpleDisplay;
+  component = InsightDisplay.SimpleChart;
```
dfarooq610, line 165: Wait is this a chart tho? I thought the way we query this is we get the literal average wait time of all questions (with filters applied), therefore it should be a SimpleDisplay

dfarooq610 (approved):
It seems pretty good overall, I left a few minor comments but damn this is so cool and the progress on this is HYPE

packages/app/pages/course/[cid]/insights.tsx
Comment (willstenzel) [reply packages/app/pages/course/[cid]/insights.tsx:42]: s -> smallInsights
d -> defaultInsights
i -> insight

Comment (willstenzel) [reply packages/app/pages/course/[cid]/insights.tsx]: I just renamed for clarity. https://github.com/sandboxnu/office-hours/pull/522/commits/bb5f7e1394c9911f497359387e661de02dc1ed86

packages/server/src/insights/insights.controller.ts
willstenzel: You're right, that could have worked in the previous changes. 

Since the response was not being used anywhere I decided to have it return void instead.

packages/server/src/insights/insight-classes.ts
willstenzel: Good catch, I just fixed that.

PR #523, opened 2021-01-14
logging

NEUDitao:


packages/server/src/login/login.controller.ts
```diff
@@ -42,6 +43,7 @@ export class LoginController {
       );
       if (!verifySignature) {
         apm.captureError('Invalid request signature');
+        Sentry.captureMessage('Invalid request signature');
```
liustanley, line 46: not sure what `parsedRequest` is but should we log it in the message here?

liustanley (approved):

packages/server/src/login/login.controller.ts
NEUDitao: sure

willstenzel (approved):

PR #527, opened 2021-01-17
Professor Office Hours fix since Async Map doesn't work as I thought it did and recurring events aren't everything

NEUDitao:


dfarooq610 (approved):
LGTM

willstenzel (approved):

packages/server/src/course/ical.service.ts
```diff
@@ -182,31 +182,33 @@ export class IcalService {
...
-        );
-        return OfficeHourModel.create({
-          queueId: professorQueue.id,
-          ...oh,
+    const processedProfessorOfficeHours = [];
+
+    for (const poh of professorOfficeHours) {
+      const professorLocation = poh.title;
+      if (!professorQueues.some((q) => q.room === professorLocation)) {
+        const newProfQ = QueueModel.create({
```
liustanley, line 190: do we need to `await` on QueueModel.create?

NEUDitao: no, the `.save` at the end fixes it

PR #529, opened 2021-01-19
Fix professor login

NEUDitao:


packages/server/src/login/login-course.service.ts
```diff
@@ -51,23 +46,21 @@ export class LoginCourseService {
...
-          );
-          userCourses.push(taCourse);
-        }
-      }),
-    );
+      for (const courseMapping of courseMappings) {
+        const taCourse = await this.courseToUserCourse(
+          user.id,
+          courseMapping.courseId,
+          c.instructor === 1 ? Role.PROFESSOR : Role.TA,
```
liustanley, line 59: Can we also change the type in `KhouryTACourse` to `number`?

packages/server/test/login.integration.ts
```diff
@@ -286,7 +286,7 @@ describe('Login Integration', () => {
       expect(ta.courses).toHaveLength(3);
     });
 
-    it('handles professor courses correctly', async () => {
+    it.only('handles professor courses correctly', async () => {
```
liustanley, line 289: Remove `.only`

liustanley (changes requested):

packages/server/src/login/login-course.service.ts
liustanley: Just checked and cypress is failing because of this too

NEUDitao: wait lol I just forgot to cmmit it HAHA

packages/server/test/login.integration.ts
NEUDitao: yes sir

willstenzel (approved):

liustanley (approved):

PR #562, opened 2021-01-29
Husky emails to northeastern backfill

IrisLiu-00:
Backfill to convert husky emails to northeastern emails, and updates to backend so that only neu emails are stored when new users login through khoury.


liustanley (approved):
Clean!!

willstenzel (approved):

PR #603, opened 2021-02-23
El il text area

### LINKED ISSUE #398: Allow queue notes to support links and possibly markdown
**Story Spec**
- [ ] Edit queue notes -> textarea
- [ ] New lines and links will display correctly in the queue notes

On multiple occasions I've seen TA's have added links in the queue notes. It would be nice to allow students to click directly on the live links instead of having them copy paste it. 

Supporting markdown would solve this issue but it would also mean that we might have to redesign the way that the queue notes are shown on the `OpenQueueCard`.

IrisLiu-00:
# Description

Allow instructors to input new lines and links in queue notes, and have new lines display properly and links be clickable. 

Also has a fix for seeding data, where new users were being seeded with bad IDs. 

Closes #398 

## Type of change

- [x] This requires a run of `yarn install`

# How Has This Been Tested?

- [x] Manually - create queue notes and put in some links and new lines
- [x] Manually - seed data and make sure it works

# Checklist:

### DIFF
--- a/packages/server/src/seed/seed.controller.ts
+++ b/packages/server/src/seed/seed.controller.ts
@@ -110,7 +110,7 @@ export class SeedController {
         lastName: 'Liu',
         photoURL:
           'https://ca.slack-edge.com/TE565NU79-UR20CG36E-cf0f375252bd-512',
-      }).save();
+      });
       await UserCourseFactory.create({
         user: user1,
         role: Role.STUDENT,
--- a/packages/server/src/seed/seed.controller.ts
+++ b/packages/server/src/seed/seed.controller.ts
@@ -3,7 +3,7 @@ import { Body, Controller, Get, Post, UseGuards } from '@nestjs/common';
 import { EventModel } from 'profile/event-model.entity';
 import { UserCourseModel } from 'profile/user-course.entity';
 import { UserModel } from 'profile/user.entity';
-import { Connection } from 'typeorm';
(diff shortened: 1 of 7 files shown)


willstenzel (commented):
Is it possible for the links to open up in a new tab?

packages/server/src/seed/seed.controller.ts
```diff
@@ -37,6 +37,8 @@ export class SeedController {
     await this.seedService.deleteAll(EventModel);
     await this.seedService.deleteAll(UserModel);
     await this.seedService.deleteAll(CourseModel);
+    const manager = getManager();
+    manager.query('ALTER SEQUENCE user_model_id_seq RESTART WITH 1;');
```
dfarooq610, line 41: What does this do?

dfarooq610 (approved):
I agree with Will that the links could open up in a new tab, but overall, this is awesome!

packages/server/src/seed/seed.controller.ts
IrisLiu-00: We're deleting all the UserModels above, but when you go back to re-seed data the UserModel ids would not start back off at 1. We have dev urls that depend on those user ids though, so we have to run a query to reset the ids since they don't automatically reset after deleting.

willstenzel (approved):
Great stuff

PR #627, opened 2021-03-07
Data insights backend cleanup

willstenzel:
Cleans up the code for the backend of data insights. 

Changes include:
* Renaming variables, file, functions for readability 
* Removing unneeded code
* Adding types when missing

(See commit history for full list of changes)


packages/server/src/insights/insights.service.ts
```diff
@@ -1,42 +1,27 @@
...
 export class InsightsService {
   constructor(private connection: Connection) {}
 
-  // Generate the output data for an insight by calling its compute function
-  async generateOutput({ insight, filters }): Promise<any> {
+  // Compute the output data for an insight and add it to the insight response
+  async computeOutput({
+    insight,
+    filters,
+  }: GenerateOutputParams): Promise<any> {
```
dfarooq610, line 20: can this be a Promise<OutputTypes>? Or whatever it was Called in common

dfarooq610 (approved):
These changes are all good. I do wonder if there is a way to better standardize these queries. 

Also, if we eventually do maneuver into a  fixed dashboard, maybe these insight classes become individual endpoints on the controller?

packages/server/src/insights/insights.service.ts
willstenzel: Yeah just made it an `Insight` type. Good catch

PR #634, opened 2021-03-12
Data insights UI/UX improvements

willstenzel:
This PR cleans up the data insights UI and make the UX more clear. Shoutout @liustanley for the help on this one. 

Changes include: 
* Reformatting the insight display options drawer
* Renaming "Open Insights Display Options" to "Edit Insights"
* Moving the RangePicker and Edit Insights button to the same line as the insights header
* Adding a tooltip that explains the data being shown when tooltip is empty 

Old
Screen Shot 2021-03-12 at 12 48 29 AM

New
Screen Shot 2021-03-12 at 12 47 37 AM


packages/app/pages/course/[cid]/insights.tsx
```diff
@@ -72,22 +71,47 @@ export default function Insights(): ReactElement {
...
-            <RangePicker
-              onChange={(_, dateString) =>
-                setDateRange({ start: dateString[0], end: dateString[1] })
-              }
-            />
-          </div>
-        </InsightContainer>
-        <Divider style={{ margin: "0 0 8px 0" }} />
-
+        <Row
```
Comment (dfarooq610) [inline packages/app/pages/course/[cid]/insights.tsx:74]: I don't really know our exacts on styled components vs inline styles, maybe these should be styled components?? Not sure

dfarooq610 (approved):

packages/app/pages/course/[cid]/insights.tsx
Comment (willstenzel) [reply packages/app/pages/course/[cid]/insights.tsx:74]: Stanley and I originally tried styled components then decided on using the ant-design row cause turned out to work better

packages/app/components/Insights/components/InsightsDisplayOptions.tsx
```diff
@@ -1,9 +1,18 @@
...
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  font-size: 18px;
+  align-items: center;
+  margin-bottom: 16px;
+`;
 interface InsightsDisplayOptionsProps {
   toggleInsightOn: (insightName: string) => void;
   toggleInsightOff: (insightName: string) => void;
```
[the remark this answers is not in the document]
willstenzel: Good catch! Previously the toggleOff function was used to be able to toggle off an insight from the hide button on the insight itself but that made for a confusing UX so I removed it. That means they no longer need to be passed in as props. 

I just moved both the `toggleInsightOn` and `toggleInsightOff` functions into the InsightsDisplayOptions component instead of passing them in as props.

PR #635, opened 2021-03-14
Create anonymize prod script

### LINKED ISSUE #620: Seed data to test data insights dashboard
**Milestone Goal: 3/11**

willstenzel:
# Description

Wrote up a script to anonymize the emails, phone numbers, photoURLs, firstName and lastName of userModels in a sql dump file. This will allow us to test data insights using real data on our local machines. 

To be able to anonymize staging data, you need to: 
1. run pg_dump to create the prod.sql file
2. run the python script to anonymize the data, which is then written into a anonymized_prod.sql file. 
3. delete the original prod.sql file. 
4. scp the new anonymized_prod sql dump. 

In the future, we may write a bash script to do these steps in one command. 

Closes #620

# How Has This Been Tested?

Please describe how you tested this PR (both manually and wtih tests)

- [ ] Ran the script using a sql dump file from staging, and manually checked the result. 

# Checklist:

### DIFF
--- a/fetch_prod.sh
+++ b/fetch_prod.sh
@@ -6,4 +6,4 @@ user=$1
 
 cat ./infrastructure/prod/anonymize.sh | ssh -J $user@login.ccs.neu.edu $user@amato.ccs.neu.edu
 sleep 5
-scp $user@amato.ccs.neu.edu:/var/www/source/infrastructure/prod/anonomized_prod.sql ./
+scp $user@amato.ccs.neu.edu:/var/www/source/infrastructure/prod/anonymized_prod.sql ./
--- a/infrastructure/prod/anonymize_sql.py
+++ b/infrastructure/prod/anonymize_sql.py
@@ -1,0 +1,30 @@
+prod = open('prod.sql', 'r') 
+anonomized_prod = open('anonomized_prod.sql', 'w')
+
+lines = prod.readlines()
+
+anonymizing = False
+
+for line in lines:
+  newLine = line
+
+  if "\.\n" == line:
+    anonymizing = False
+  
+  if anonymizing:
+    elements = line.split()
+    id = elements[0]
+    elements[1] = id + '@northeastern.edu'
+    elements[2] = ''
+    elements[5] = 'user' + id
+    elements[6] = 'person' + id
+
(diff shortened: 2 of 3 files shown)


NEUDitao:
even without a bash command, wouldn't Python's `os.system()` allow you to run shell commands? and then `os.remove()` exists too iirc

willstenzel:
@NEUDitao good call. I just added the `os` commands, let us know how that looks

liustanley (changes requested):
Looks great, I'm gonna work on making a script to also `scp`. Only thing left is that phone numbers are still left in the anonymized sql file, and should all be removed

fetch_prod.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# Fetches anonymized prod dump for development seeding
+set -ex
+
+user=$1
+
+cat ./infrastructure/prod/anonymize.sh | ssh -J $user@login.ccs.neu.edu $user@amato.ccs.neu.edu
```
NEUDitao, line 7: AFAIK bash is supposed to wait for the `ssh` session to be over before continuing w/ the script. instead of using `cat`, might be better to just move `./infrastrucutre....` to the RHS of `ssh`

willstenzel:
> Looks great, I'm gonna work on making a script to also `scp`. Only thing left is that phone numbers are still left in the anonymized sql file, and should all be removed

Good call, I just added to the script so all phone numbers will be removed

NEUDitao (approved):
i'm hip

infrastructure/prod/anonymize_sql.py
```diff
@@ -0,0 +1,48 @@
...
+
+anonymizing_user = False
+anonymizing_phone_numbers = False
+
+lines = prod.readlines()
+
+for line in lines:
+  newLine = line
+
+
```
dfarooq610, line 16: Is it just the device I am looking at this on, or are these if statements not indented?

liustanley (approved):

liustanley:
I think we're good to merge, we weren't able to fully test because the `officehours` group isn't allowed in the postgres database yet, but everything should be good to go after we get this unblocked!

Issue #621, opened 2021-03-04
Implement TA question grouping


liustanley:
**Milestone Goal: 3/18**
- [ ] Support grouping in frontend (3/18)
- [ ] @ofloody Support grouping in backend (3/25)

Mockups: https://www.figma.com/file/y0vUWL5G2664qsG6eUq9PQ/New-Features?node-id=107%3A0

Spec (needs to be revised): https://www.notion.so/sandboxnu/Plan-af4426868ac64e3ca01234fe1877fc5c

PR #639, opened 2021-03-16
WIP: Question Grouping Frontend

IrisLiu-00:
# Description

Implements the frontend for question grouping (issue #621)  
This is a WIP so there's still a lot of TODOs in the code for hooking things up with the backend. 

## Type of change

# How Has This Been Tested?

- [ ] Manually: hit `Add Students` under the question grouping sidebar, add some students, then hit the call button to get the second view

(Students currently in a group do not show up on the sidebar, because the group is not actually created on the backend and would need to be passed into the `TAQueueListSection`)

# Checklist:


liustanley:
Looks great! My only comments are:
- [ ] It would be nice if the blue highlight and the bottom line when you select someone for grouping extended all the way, but if this turns out to be ridiculous to implement it's all good
- [ ] This is my bad because I forgot to include it in the story mockups, but we want an additional form option when the student is creating/editing a question. The form option should be "Would you like the option of being helped in a group session?", with Yes and No options. It can also have a description saying "Clicking Yes may result in a shorter wait time if others have the same question as you."

Since I forgot to include the second one, it's ok if you want to wait until next sprint to implement it

liustanley:
Oh one more thing:
- [ ] Instead of the "Remove from Queue" option below, we want the "Requeue" and "Can't Find" buttons

Screen Shot 2021-03-17 at 7 18 55 PM

IrisLiu-00:
> * [ ]  Instead of the "Remove from Queue" option below, we want the "Requeue" and "Can't Find" buttons

Someone can correct me if I'm wrong on this, but I believe that will have the proper buttons once the backend is hooked up - I'm using the `TAQueueDetailButtons`, which will show different buttons depending on the question status. I didn't do any backend calls to change questions statuses when you hit the call button, although maybe I should have put something 🤔 

> * [ ]  It would be nice if the blue highlight and the bottom line when you select someone for grouping extended all the way, but if this turns out to be ridiculous to implement it's all good

Yeah I didn't do this at first because I thought it would be nasty but it's not all that nasty actually. Should be fixed now

> * [ ] This is my bad because I forgot to include it in the story mockups, but we want an additional form option when the student is creating/editing a question. The form option should be "Would you like the option of being helped in a group session?", with Yes and No options. It can also have a description saying "Clicking Yes may result in a shorter wait time if others have the same question as you."

Oops we will do this, although it may take a few extra days before I get around to it !!

liustanley (approved):

willstenzel:
> Oh one more thing:
> 
> * [ ]  Instead of the "Remove from Queue" option below, we want the "Requeue" and "Can't Find" buttons
> 
> Screen Shot 2021-03-17 at 7 18 55 PM

@liustanley does it make sense to have all three buttons ("Remove from Queue", "Requeue", "Can't Find") here? Or is there a reason to only have the "Requeue" and "Can't Find" buttons?

PR #641, opened 2021-03-18
Requestion

### LINKED ISSUE #585: Alerts table and endpoint
**Milestone Goal: 3/7**

Should push alerts and modals from the backend
Also integrate SSE into this

- [ ] Alerts SWR lives in the NavBar or Container
- [ ] Add an endpoint for Alerts
- [ ] On an Alert, display a modal to the Student that they have to dismiss
- [ ] Clicking on the Alert would take you to the queue page to do the required action
- [ ] Alert table: `studentId, alertType, course`
- [ ] On dismiss, delete the Alert from the table

ofloody:
# Description

Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context. 
List any dependencies that are required for this change.

Closes #585 

## Type of change

# How Has This Been Tested?

- [ ] Manual Testing. Have TA try to requeue a student's question


packages/app/components/Queue/TA/TAQueueListItem.tsx
```diff
@@ -19,6 +21,22 @@ const Container = styled.div<{ selected: boolean }>`
   align-items: flex-start;
 
   padding-top: 16px;
+  return string;
```
willstenzel, line 24: Don't think this is meant to be here

packages/app/components/Queue/TA/TAQueueListItem.tsx
```diff
@@ -19,6 +21,22 @@ const Container = styled.div<{ selected: boolean }>`
...
+}
+
+const Container = styled.div<{ selected: boolean }>\`
```
willstenzel, line 27: On my machine it's not liking the `\` before the back tick. Also this `Container` is overriding the one on line 19

```suggestion
const Container = styled.div<{ selected: boolean }>`
```

packages/api-client/index.ts
```diff
@@ -168,6 +171,17 @@ class APIClient {
     get: async (): Promise<GetReleaseNotesResponse> =>
       this.req("GET", `/api/v1/release_notes`),
   };
+  alerts = {
+    get: async (courseId: number): Promise<GetAlertsResponse> =>
+      this.req("GET", `/api/v1/alerts/${courseId}`),
+    create: async (params: CreateAlertParams): Promise<CreateAlertResponse> =>
+      this.req("POST", `/api/v1/alerts`, CreateAlertResponse, params),
+    close: async (alertId: number): Promise<void> => {
+      console.log("balls", alertId);
```
dfarooq610, line 180: Balls

packages/server/src/alerts/alerts.service.ts
```diff
@@ -0,0 +1,35 @@
...
+    for (const alert of alerts) {
+      // Might be one of the few usecases for ReasonML
+
+      switch (alert.alertType) {
+        case AlertType.REPHRASE_QUESTION:
+          const payload = alert.payload as RephraseQuestionPayload;
+          const question = await QuestionModel.findOne(payload.questionId);
+          // QUESTION: is there a better way of detecting if the question is stale due to
+          // the queue being closed and then re-opened? I think queue cleaning will
+          // solve this issue, but... who knows?
```
willstenzel, line 23: This should work alright since all stale questions will be cleaned each day

packages/server/src/alerts/alerts.service.ts
```diff
@@ -0,0 +1,35 @@
...
+          const queue = await QueueModel.findOne(payload.queueId);
+          if (question.closedAt || !(await queue.checkIsOpen())) {
+            console.log(`Rephrase Question alert with id ${alert.id} expired`);
```
willstenzel, line 26: Do we want to mark the alert as resolved here so it doesn't get keep getting pulled when we make the db query for the alerts?

packages/server/src/alerts/alerts.entity.ts
```diff
@@ -0,0 +1,46 @@
...
+  @Exclude()
+  userId: number;
+
+  @ManyToOne((type) => CourseModel, (course) => course.alerts)
+  @JoinColumn({ name: 'courseId' })
+  course: CourseModel;
+
+  @Column({ nullable: true })
+  @Exclude()
+  courseId: number;
```
willstenzel, line 42: Why do we need a course relation for an alert?

willstenzel (changes requested):

liustanley:
@ofloody I've only had a chance to look at the feature by running the code, and I caught some issues that need to be addressed before further reviewing:
- [ ] After you finish editing your question, the Alert is never dismissed (the student can never leave the alert). Maybe once the student clicks "Edit Question", it will also send a request to `/alerts` to delete their alert?
- [ ] Clicking "Edit Question" shouldn't refresh the page, ideally the flow would dismiss the alert from the screen, and then smoothly bring up the Edit Question Modal
- [ ] I also encountered an issue when spamming the requestion button on the TA side, the student side background becomes black (possibly from duplicate alerts), but duplicate alerts should not be possible

Screen Shot 2021-03-19 at 1 01 09 PM

packages/app/components/Queue/Student/StudentQuestionRephraseModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <Modal
+      visible={true}
+      closable={false}
+      footer={[
+        <Button
+          type={"primary"}
+          key={"continue"}
+          href={`/course/${courseId}/queue/${payload.queueId}?edit_question=true`}
+          onClick={handleClose}
```
liustanley, line 25: Something isn't working with `handleClose`, does onClick behave weirdly with async functions? Or is there a problem with the API request?

packages/app/components/Queue/Student/StudentQuestionRephraseModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+}: StudentQuestionRephraseModalProps): ReactElement {
+  return (
+    <Modal
+      visible={true}
+      closable={false}
+      footer={[
+        <Button
+          type={"primary"}
+          key={"continue"}
+          href={`/course/${courseId}/queue/${payload.queueId}?edit_question=true`}
```
liustanley, line 24: I like this solution, but I'd prefer if it didn't refresh the entire page. Maybe instead of an `href` here, we could wrap the Button with a [Link](https://nextjs.org/docs/api-reference/next/link), see suggestion above

packages/app/components/Queue/Student/StudentQuestionRephraseModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  courseId,
+  payload,
+  handleClose,
+}: StudentQuestionRephraseModalProps): ReactElement {
+  return (
+    <Modal
+      visible={true}
+      closable={false}
+      footer={[
+        <Button
```
liustanley, line 21:
```suggestion
        <Link 
            href={{ pathname: "/course/[cid]/queue/[qid]" query: { edit_question: 'true' } }} 
            as={`/course/${courseId}/queue/${payload.queueId}`}
            >
            <Button
```

packages/server/src/alerts/alerts.controller.ts
```diff
@@ -0,0 +1,103 @@
...
+        alertType,
+        userId: targetUserId,
+        resolved: null,
+      },
+    });
+
+    /*
+
+    // If the same user already has an alert for this then don't create a new one
+    if (anotherAlert) {
```
liustanley, line 67: This code section should be uncommented to prevent duplicate alerts, but make sure the error doesn't propagate to the frontend.

liustanley (changes requested):

packages/server/src/alerts/alerts.entity.ts
ofloody: I think so because we want the alerts to be persistent in every part of the app (since it's in the nav bar). Alerts also are meant to be more generic and could apply to non queue specific things ex. cancelled office hours

packages/app/components/Queue/Student/StudentQuestionRephraseModal.tsx
ofloody: When I tried this, I didn't trigger the edit question modal

packages/app/components/Nav/AlertsContainer.tsx
```diff
@@ -0,0 +1,43 @@
+import { API } from "@koh/api-client";
+import { AlertType, RephraseQuestionPayload } from "@koh/common";
+import React, { ReactElement } from "react";
+import useSWR from "swr";
+import { useRouter } from "next/router";
+import StudentQuestionRephraseModal from "../Queue/Student/StudentQuestionRephraseModal";
+
+type AlertsContainerProps = {
+  courseId: number; // TODO: replace this with a courseId that comes from the RephraseQuestionPayload
```
liustanley, line 9: Course should match the course that the alert originated from, this courseId is coming from the current course in the navbar (not what we want)

packages/app/components/Queue/Student/StudentQuestionRephraseModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+  handleClose,
+}: StudentQuestionRephraseModalProps): ReactElement {
+  return (
+    <Modal
+      visible={true}
+      footer={[
+        <Button
+          type={"primary"}
+          key={"continue"}
+          onClick={() => handleClose(courseId, payload.queueId)} // TODO: replace courseId with courseId from payload
```
liustanley, line 23: Same thing here, course should match the course that the alert originated from, this courseId is coming from the current course in the navbar (not what we want)

liustanley (changes requested):

packages/app/components/Nav/NavBar.tsx
```diff
@@ -180,9 +181,10 @@ export default function NavBar({ courseId }: NavBarProps): ReactElement {
     });
   }
 
-  return (
+  return courseId ? (
```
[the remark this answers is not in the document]
NEUDitao: i tried but it got upsetti for some reason

packages/server/src/alerts/alerts.service.ts
NEUDitao: good call

packages/app/components/Queue/TA/TAQueueListItem.tsx
ofloody: It's called in line 101

packages/app/components/Queue/TA/TAQueueListItem.tsx
ofloody: without it the following is not properly evaluated as a string. It is weird idk if we need that section

packages/app/components/Queue/TA/TAQueueListItem.tsx
willstenzel: Sorry I should have been more specific, I was talking about the `return string` part

liustanley (approved):

packages/server/src/alerts/alerts.service.ts
NEUDitao: bet, the comment can be removed then

packages/server/src/alerts/alerts.service.ts
NEUDitao: i'll do this later

willstenzel (approved):

packages/app/components/Queue/TA/TAQueueDetailButtons.tsx
```diff
@@ -135,6 +161,15 @@ export default function TAQueueDetailButtons({
             </span>
           </Tooltip>
         </Popconfirm>
+        <Tooltip title="Ask the student to rework their question">
```
willstenzel, line 164: Should this say `"Ask the student to add more detail to their question"` so it matches with what we show the student when the button is pressed?

PR #654, opened 2021-04-05
small insights bug fixes

dfarooq610:
# Description

If there are no questions answered, the median helping/wait time is 0. 
If the network request containing an insight output is falsy, but still a value like 0, we want to render it. We only want the loading shell with undefined values. 

# How Has This Been Tested?

- [ ] Clicking around forcefully

# Checklist:

### DIFF
--- a/packages/app/pages/course/[cid]/insights.tsx
+++ b/packages/app/pages/course/[cid]/insights.tsx
@@ -207,7 +207,7 @@ function RenderInsight({
       }
     >
       {/* {true ? <Spin style={{ margin: '20% 45%' }} /> : <DataComponent key={insightName} output={insightOutput} />} */}
-      {!insightOutput ? (
+      {insightOutput === undefined ? (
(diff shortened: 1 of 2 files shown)


packages/app/pages/course/[cid]/insights.tsx
```diff
@@ -207,7 +207,7 @@ function RenderInsight({
       }
     >
       {/* {true ? <Spin style={{ margin: '20% 45%' }} /> : <DataComponent key={insightName} output={insightOutput} />} */}
```
Comment (liustanley) [inline packages/app/pages/course/[cid]/insights.tsx:209]: Should we remove this comment?

liustanley (approved):

packages/app/pages/course/[cid]/insights.tsx
Comment (dfarooq610) [reply packages/app/pages/course/[cid]/insights.tsx:209]: @willstenzel

Comment (willstenzel) [reply packages/app/pages/course/[cid]/insights.tsx:209]: yeah my b

PR #672, opened 2021-04-24
alerts save...

### LINKED ISSUE #661: Fix rephrase question bug
The feature is temporarily disabled, but there's a problem with getting alerts for a user.

The alert is created correctly in the table, but something is messing up the response (might be the stale alerts clean function).

NEUDitao:
# Description

Closes #661

## Type of change

# How Has This Been Tested?

- [x] Manually
- [x] unit test

# Checklist:


packages/server/src/alerts/alerts.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+          queueId: 1,
+          questionId: 420,
+        }),
+      ).toBeTruthy();
+    });
+
+    it('incorrect rephrase question payloads fail', () => {
+      expect(
+        service.assertPayloadType(AlertType.REPHRASE_QUESTION, {
+          courseId: 'PYHUGYHIF',
```
NEUDitao, line 102: @liustanley

packages/server/src/alerts/alerts.service.ts
```diff
@@ -20,17 +26,42 @@ export class AlertsService {
...
+          if (question.closedAt || !isQueueOpen) {
+            console.log(
+              `Rephrase Question alert with id ${
+                alert.id
+              } expired ${isQueueOpen} ${JSON.stringify(question)}`,
+            );
             if (!question.closedAt) {
               question.closedAt = new Date();
+              await question.save();
             }
```
willstenzel, line 39: This is essentially closing a question that is stale right? 

If so, it make more senes to leave all this to the `queue-clean.service` so that we can keep track of what's happening where. 

 If you want to do it here, that's alright but you'll probably also want to change the question status to `Stale`

PR #731, opened 2021-10-24
Hide data insights and heatmap

### LINKED ISSUE #725: [Queue Refactor] - Disable data insights and heatmap
** do this on the `hybrid-queues` branch please!!

As we refactor the QueueModel, it's going to break our data insights and heatmap. We're gonna disable this stuff until 

Frontend:
- hide the data insights page and insights tab
- hide the heatmap

Backend:
- don't return heatmap/insights data (I'm not sure if you really need to disable the insights endpoints much, but you will need to avoid returning the heatmap with the rest of the GET course data. There might be some more investigation here on exactly how insights data is returned from the backend)

brianyyang:
# Description

As we refactor QueueModel, insights and heatmap might break so we disabled those on the front end and back end.

Closes #725 

## Type of change

# How Has This Been Tested?

We didn't the insights.services file so the insights test suite still works properly. Instead we throw errors if anyone tries to reach the endpoints through the controller.

As for the front end, make sure the insights tab on the nav bar doesn't show up for professors. Also, if you manually navigate to the page through ./course/#/insights, make sure no back end errors are popping up and the page should tell you that insights are disabled.

# Checklist:


IrisLiu-00 (approved):
LGTM! Although I will defer to @dfarooq610 since he knows much more about data insights than I

packages/app/components/Nav/NavBar.tsx
```diff
@@ -180,13 +180,16 @@ export default function NavBar({ courseId }: NavBarProps): ReactElement {
       text: "Queue",
     });
   }
+
+  /**
```
dfarooq610, line 184: Nit: just add a TODO here so we know to re-enable it eventually (maybe link this issue on GH)

packages/app/pages/course/[cid]/insights.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { ReactElement, useState } from "react";
+import React, { ReactElement } from "react";
```
Comment (dfarooq610) [inline packages/app/pages/course/[cid]/insights.tsx:1]: Instead of commenting this all out, can we just redirect to a 404 for a bit lmao?idk

dfarooq610 (changes requested):
Something else you might want to disable is the generation of the heatmap. The heatmap service is within the course, and the easiest way to "disable" it is to just set it to the Boolean False when you get a course

packages/server/test/course.integration.ts
```diff
@@ -401,7 +401,7 @@ describe('Course Integration', () => {
       });
 
       const thenThreeMoreHours = new Date(yesterdayPlusTwoHours);
-      thenThreeMoreHours.setUTCHours(yesterday.getUTCHours() + 3);
+      thenThreeMoreHours.setUTCHours(yesterdayPlusTwoHours.getUTCHours() + 3);
```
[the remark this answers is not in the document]
brianyyang: yes

packages/app/components/Nav/NavBar.tsx
brianyyang: good idea!

packages/app/pages/course/[cid]/insights.tsx
Comment (brianyyang) [reply packages/app/pages/course/[cid]/insights.tsx:1]: that sounds like a plan

Comment (brianyyang) [reply packages/app/pages/course/[cid]/insights.tsx:1]: nevermind, we decided having the message saying insights is temporarily disabled would be a nice to have

dfarooq610 (approved):

PR #740, opened 2021-11-01
Prevent queues from opening automatically

### LINKED ISSUE #729: [Queue Refactor] -- Refactor Queues to Not Open Automatically
Look into how the isOpen field works on a queue: right now students are able to join a queue before the queue is scheduled to happen. What we want to do is prevent queues from being open automatically. 

TODO: 
- [x] Do some research about how queues open, and how students can join queues within our application.
- [x] Remove the functionality that allows queues to open automatically.
- [ ] Perhaps write a migration to remove the isOpen field on a queue if not needed

ashdawngary:
# Description

Updates queues' open status to be only based on whether there are staff checked in or not (`QueueModel.staffList` is not empty). Function `checkIsOpen` is still async, but should resolve nearly instantly since it is returning a promise that is resolved instantly.

Side Effects: queues with no staff checked in are not present anymore -- ie invisible for both the TA and the student.  That being said, TA's still have a check-in button to get into the queue.  Professor queues are immune and still visible.

Tests that relied on the old defintion of 'open' were updated (`queue.integration.ts`, `course.integration.ts`, `question.integration.ts` ) to make sure that queue is open when applicable

Closes #729 

## Type of change

# How Has This Been Tested?

- [ ] Using two browsers, log in as the student and the TA, and see how the queue's visibility changes based on whether TA is checked in or not.
- [ ] Automatic Tests: 2 tests in `queue.integration.ts`: first has oh in schedule, but no TA checked in (queue should be closed), and one where a TA is present (should be open).

# Checklist:

### DIFF
--- a/packages/server/src/queue/queue.entity.ts
+++ b/packages/server/src/queue/queue.entity.ts
@@ -71,12 +71,7 @@ export class QueueModel extends BaseEntity {
   isOpen: boolean;
 
   async checkIsOpen(): Promise<boolean> {
-    if (this.staffList && this.staffList.length > 0) {
-      this.isOpen = true;
-      return true;
-    }
-
-    this.isOpen = await this.areThereOfficeHoursRightNow();
+    this.isOpen = this.staffList && this.staffList.length > 0;
     return this.isOpen;
   }
 
--- a/packages/server/test/question.integration.ts
+++ b/packages/server/test/question.integration.ts
@@ -15,6 +15,7 @@ import { QuestionModule } from '../src/question/question.module';
 import {
   ClosedOfficeHourFactory,
   CourseFactory,
+  OfficeHourFactory,
   QuestionFactory,
   QuestionGroupFactory,
   QueueFactory,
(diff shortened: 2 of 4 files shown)


packages/server/test/question.integration.ts
```diff
@@ -35,23 +35,31 @@ describe('Question Integration', () => {
...
-
-      await UserCourseFactory.create({
-        user: await UserFactory.create(),
-        course: course,
+        queue: queue,
+        creator: student1.user,
       });
 
-      const response = await supertest({ userId: 99 })
+      const response2 = await supertest({ userId: 99 })
```
IrisLiu-00, line 58: Does this need to be named response2?

packages/server/test/question.integration.ts
```diff
@@ -35,23 +35,31 @@ describe('Question Integration', () => {
   describe('GET /questions/:id', () => {
     it('gets a question with the given id', async () => {
       const course = await CourseFactory.create();
+      const student1 = await UserCourseFactory.create({
```
IrisLiu-00, line 38: out of curiosity what's the reason for adding student1 and manually passing in the question creator?

packages/server/src/queue/queue.entity.ts
```diff
@@ -71,12 +71,7 @@ export class QueueModel extends BaseEntity {
   isOpen: boolean;
```
IrisLiu-00, line 72: (this comment isnt related to this line, I'm just putting it here so we can thread responses)
It looks like queues with no staff members now no longer show up for professors on their "check in" modal (the one where they can pick which queue to check into). Can you look into why that's happening 🤔 

Also not sure if this would interfere with @VKong6019 and @tiingweii-shii with their work on extending ad hoc queues to TAs, although you guys are only on the TA side of things right

IrisLiu-00 (changes requested):
Thanks for updating all those tests 🙏

packages/server/test/question.integration.ts
ashdawngary: in order to agree with the past snaposhot, i needed the person with ID 1 to submit the question -- but If I create a TA afterwards, then the sender has id 3(unless i specify the sender manually).  I could update the snapshot but this seemed easier to be consistent

VKong6019 (approved):
hehe

PR #752, opened 2021-11-05
Upgrade Next to v11

IrisLiu-00:
# Description
Update version of next from 10 to 11.
* I don't really understand most of the changes in this PR lmao i just threw shit at it until things ran
* Delete usages of `<Container/>` because that's deprecated
* It seems like TypeORM somehow changed through this? changed some integration tests and a migration got generated for some reason
* Switch cypress github action to run on chrome instead of electron, which seems to stop it from hanging for 1hr +

## Type of change

- [x] This requires a run of `yarn install`

# Checklist:


packages/server/test/course.integration.ts
```diff
@@ -298,11 +298,8 @@ describe('Course Integration', () => {
         staffList: [],
       });
 
-      const UCM = UserCourseModel.findOne({
-        where: { userId: ta.id, courseId: queue.courseId },
-      });
       const events = await EventModel.find({
-        where: { user: UCM },
+        where: { userId: ta.id },
```
IrisLiu-00, line 302: I'm fairly sure this test was originally incorrectly written - it should be finding Events with the given User, not the given UserCourseModel. It started failing as written on this branch - maybe TypeORM changed how it does joins?

Making this same switch on master caused the test to fail as well, because the course controller is actually logging a "checkout" event even if the TA wasn't originally in the queue. I've updated the course controller to not log an event.

dfarooq610 (approved):
I launched it locally and did some basic E2E testing (creating questions as a student then editing it, checking in as a TA and creating a group, updating my user settings, etc). 

Everything seemed pretty good, although it didn't seem super performant. Maybe that is my local device? Seeding data also took a bit. 

That aside, it's probably good to go. Feel free to merge during the meeting

IrisLiu-00:
Uhh dang yeah it was also slower on my computer.. but ok thanks for looking i will merge 👍

PR #755, opened 2021-11-09
Add Disabled Field to queues

### LINKED ISSUE #746: [Queue Refactor] -- Enable disabling queues from the backend
We want to eventually be able to disable queues. 
- add a `disabled` attribute to queues, default `false` (this will need a migration and a backfill)
- create an endpoint to disable a queue (permissioned to TAs and Profs)
- this endpoint should also mark all the questions in the queue as QuestionStatus.stale

ashdawngary:
# Description

Added disabled field to queues.  The field is a Boolean and intended to hide queues forever from the frontend (while `isOpen` is for whether a queue is joinable by a student.)  The default when a queue opens is false, but if a user with the proper role (TA/Prof for TA queues, Prof for prof queues), sends a delete request to `/queues/:id`, then it will disable the queue in the backend.

Also added a migration for preexisting queues in `backfill-queue-isdisabled.ts`

Lastly added tests in `queue.integration.ts` for the endpoint created and different situations.  Tested all my changes with cypress locally.

Closes #746 

## Type of change

# How Has This Been Tested?

- [ ] New `describe` test group within `queues.integration.ts` titled `DELETE /queues/:id` with tests for disabling queues, nuances (ex: ta cant close prof queues, students can close queues in general).  Also test invalid queues.

Not testable from the frontend yet, only created the endpoint and the proper modifications to the database 

# Checklist:


packages/server/src/queue/queue.controller.ts
```diff
@@ -144,4 +145,39 @@ export class QueueController {
...
+      }
+    }
+
+    queue.isDisabled = true;
+
+    try {
+      // try to save queue
+      await queue.save();
+    } catch (err) {
+      console.error(err);
```
dfarooq610, line 176: Can the SAVE_QUEUE error message be a part of this

dfarooq610 (commented):
It's a bit of scope creep of the ticket, but do you think we should also disallow adding questions to disabled queues? @IrisLiu-00

IrisLiu-00:
> It's a bit of scope creep of the ticket, but do you think we should also disallow adding questions to disabled queues? 

That would just be backend validation, right? Theoretically disabled queues won't show on the FE for students to add questions to.
I think it would be good to have though, it should be just a quick check in the add question endpoint right 🤔

Issue #708, opened 2021-10-09
Update question group Teams link to have default message


IrisLiu-00:
Depends on question grouping PR and default message PR

PR #767, opened 2021-11-13
Add default teams msg to question grouping

IrisLiu-00:
# Description
Adds the default teams message to Teams calls opened from a question group. 

Also tweak the logic to get the default message on the front end, so that if `includeDefaultMessage` is true but the `defaultMessage` is null, the teams call opens with no message instead of "null". I realized that this shouldn't actually happen if we do the backfill to set the defaultMessage of "hi my name is..." on everyone, but it's probably good to have the extra safety anyways. 

Closes https://github.com/sandboxnu/office-hours/issues/708

## Type of change

# How Has This Been Tested?

- Manual test: Set a default teams message and open a question group
- Manual test: When you delete and seed data, you can get into a state where you have the default message turned on but the message itself is empty. Verify that calling someone in this state does not autofill the teams chat. 

# Checklist:

### DIFF
--- a/.github/scripts/isMigrationNeeded.sh
+++ b/.github/scripts/isMigrationNeeded.sh
@@ -6,7 +6,6 @@ set -e
 SUB='No changes in database schema were found'
 {
   log=$(yarn migration:generate -n CHECK 2>&1)
-  echo "$log"
 } || {
   echo "$log"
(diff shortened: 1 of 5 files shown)


nicoledanuwidjaja (approved):
Thank you 💯 🙏

tiingweii-shii (approved):

PR #773, opened 2021-11-21
Update KhouryDataParams course types

### LINKED ISSUE #743: [Login Flow] -- Update User Course Registration
- [x] Update the Khoury Data Export types used in the login controller. Updated types can be found in Juan's doc. This is for Professors, TAs and Students. 
- [x] The Course Section Mapping Model needs to be updated. Replace `Generic Course Name` with `CRN`. 
- [x] Update the Login Course Service: update particularly how we create user courses using these new types.

nicoledanuwidjaja:
# Description
As per new login flow, updated course types in `KhouryDataParams` for new `KhouryCourse` and `KhouryProfCourse` objects. New course types now support courses identified by `crn`.

This PR will break existing data flow!

Closes #743 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

Updated tests in `login.integration.ts`.

# Checklist:

### DIFF
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -15,12 +15,12 @@ import {
   GetQueueResponse,
   GetReleaseNotesResponse,
   GroupQuestionsParams,
-  //GetSelfEnrollResponse,
+  GetSelfEnrollResponse,
   ListInsightsResponse,
   ListQuestionsResponse,
   SemesterPartial,
   //SubmitCourseParams,
-  //TACheckinTimesResponse,
+  TACheckinTimesResponse,
   TACheckoutResponse,
   TAUpdateStatusResponse,
   UpdateCourseOverrideBody,
@@ -114,6 +114,7 @@ class APIClient {
     /*
     submitCourse: async (params: SubmitCourseParams): Promise<void> =>
       this.req("POST", `/api/v1/courses/submit_course`, undefined, params),
+    */
     getTACheckinTimes: async (
       courseId: number,
       startDate: string,
@@ -132,7 +133,6 @@ class APIClient {
       this.req("GET", "/api/v1/self_enroll_courses"),
     createSelfEnrollOverride: async (courseId: number): Promise<void> =>
       this.req("POST", `/api/v1/create_self_enroll_override/${courseId}`),
-    */
   };
   taStatus = {
     checkIn: async (
--- a/packages/api-client/index.ts
+++ b/packages/api-client/index.ts
@@ -15,12 +15,12 @@ import {
   GetQueueResponse,
   GetReleaseNotesResponse,
   GroupQuestionsParams,
-  GetSelfEnrollResponse,
+  //GetSelfEnrollResponse,
   ListInsightsResponse,
   ListQuestionsResponse,
   SemesterPartial,
-  SubmitCourseParams,
-  TACheckinTimesResponse,
+  //SubmitCourseParams,
+  //TACheckinTimesResponse,
   TACheckoutResponse,
   TAUpdateStatusResponse,
   UpdateCourseOverrideBody,
@@ -111,6 +111,7 @@ class APIClient {
         `/api/v1/courses/${courseId}/withdraw_course`,
         undefined
       ),
+    /*
     submitCourse: async (params: SubmitCourseParams): Promise<void> =>
       this.req("POST", `/api/v1/courses/submit_course`, undefined, params),
     getTACheckinTimes: async (
@@ -131,6 +132,7 @@ class APIClient {
       this.req("GET", "/api/v1/self_enroll_courses"),
     createSelfEnrollOverride: async (courseId: number): Promise<void> =>
       this.req("POST", `/api/v1/create_self_enroll_override/${courseId}`),
+    */
   };
   taStatus = {
     checkIn: async (
--- a/packages/server/src/login/course-section-mapping.entity.ts
+++ b/packages/server/src/login/course-section-mapping.entity.ts
@@ -13,12 +13,8 @@ export class CourseSectionMappingModel extends BaseEntity {
   @PrimaryGeneratedColumn()
   id: number;
 
-  // This is the course name that is sent to us from the khoury amin backend
   @Column()
-  genericCourseName: string;
-
-  @Column()
-  section: number;
+  crn: number;
 
   // Represents the course that this maps to
   @ManyToOne((type) => CourseModel)
--- a/packages/server/src/login/login-course.service.ts
+++ b/packages/server/src/login/login-course.service.ts
@@ -1,4 +1,4 @@
-import { KhouryCourse, KhouryDataParams, Role, Season } from '@koh/common';
+import { isKhouryCourse, KhouryDataParams, Role, Season } from '@koh/common';
 import { Injectable } from '@nestjs/common';
 import { CourseModel } from 'course/course.entity';
 import { CourseSectionMappingModel } from 'login/course-section-mapping.entity';
@@ -31,11 +31,8 @@ export class LoginCourseService {
     const userCourses = [];
 
     for (const c of info.courses) {
-      if (c instanceof KhouryCourse) {
-        const course: CourseModel = await this.courseCRNToCourse(
-          c.crn,
-          c.semester,
-        );
+      if (isKhouryCourse(c)) {
+        const course = await this.courseCRNToCourse(c.crn, c.semester);
 
         if (course) {
           const userCourse = await this.courseToUserCourse(
@@ -56,12 +53,14 @@ export class LoginCourseService {
             c.semester,
           );
 
-          const profUserCourse = await this.courseToUserCourse(
-            user.id,
-            profCourse.id,
-            Role.PROFESSOR,
-          );
-          userCourses.push(profUserCourse);
+          if (profCourse) {
+            const profUserCourse = await this.courseToUserCourse(
+              user.id,
+              profCourse.id,
+              Role.PROFESSOR,
+            );
+            userCourses.push(profUserCourse);
+          }
         }
       }
     }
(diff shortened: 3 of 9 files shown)


nicoledanuwidjaja:
Cypress tests are failing:
```
Failed to compile.

./components/Settings/CourseOverrideSettings.tsx:77:30
Type error: Property 'toggleSelfEnroll' does not exist on type '{ get: (courseId: number) => Promise<GetCourseResponse>; updateCalendar: (courseId: number) => Promise<unknown>; getCourseOverrides: (courseId: number) => Promise<...>; addOverride: (courseId: number, params: UpdateCourseOverrideBody) => Promise<...>; deleteOverride: (courseId: number, params: UpdateCourseOverrideBo...'.

  75 |         <Switch
  76 |           onChange={async () => {
> 77 |             await API.course.toggleSelfEnroll(courseId);
     |                              ^
  78 |             mutateCourse();
  79 |           }}
  80 |           defaultChecked={course?.selfEnroll}

File: packages/server/src/login/course-section-mapping.entity.ts [hunk 1]
```diff
@@ -13,12 +13,8 @@ export class CourseSectionMappingModel extends BaseEntity {
   @PrimaryGeneratedColumn()
   id: number;
 
-  // This is the course name that is sent to us from the khoury amin backend
   @Column()
-  genericCourseName: string;
-
-  @Column()
-  section: number;
+  crn: number;
```
dfarooq610, line 17: Y'all need to make a migration for this

packages/server/src/login/login-course.service.ts
```diff
@@ -86,13 +86,10 @@ export class LoginCourseService {
...
 
-  public async courseSectionToCourse(
-    courseName: string,
-    courseSection: number,
-  ): Promise<CourseModel> {
+  public async courseCRNToCourse(courseCRN: number): Promise<CourseModel> {
     const courseSectionModel = (
       await CourseSectionMappingModel.find({
-        where: { genericCourseName: courseName, section: courseSection },
+        where: { crn: courseCRN },
```
dfarooq610, line 92: This works because old courses _should_ be disabled, but that is through personal maintenance (running scripts and whatnot)

@IrisLiu-00 should we validate the semester name just to double check?

dfarooq610 (changes requested):

packages/server/src/login/login-course.service.ts
```diff
@@ -31,38 +31,38 @@ export class LoginCourseService {
...
+      // loop through section groups
+      for (const c of info.prof_courses) {
+        // loop through CRNs to create courses
+        for (const courseCRN of c.crns) {
+          const courseMappings = (
+            await CourseSectionMappingModel.find({
+              where: { crn: courseCRN }, // TODO: Add semester support
+              relations: ['course'],
+            })
+          ).filter((cm) => cm.course.enabled);
```
IrisLiu-00, line 56: It looks like the logic here is the same as what's happening in `courseCrnToCourse`, so you can probably just call that function. Also I believe you only need to inspect one of the CRNs here - all of them should map to the same CourseModel since they're all for the same section group.

packages/server/src/login/login-course.service.ts
IrisLiu-00: Yeah I think this function should also take in a semester, since CRNs can potentially repeat between semesters. The MappingModel doesn't have a semester attribute but the CourseModel does - you'll have to do some joins and filters here. You'll also need to map the semester string from the admin data to our SemesterModels :P

packages/api-client/index.ts
```diff
@@ -15,12 +15,12 @@ import {
   GetQueueResponse,
   GetReleaseNotesResponse,
   GroupQuestionsParams,
-  GetSelfEnrollResponse,
+  //GetSelfEnrollResponse,
```
IrisLiu-00, line 18: I'm guessing this being commented out is why Vercel/Cypress/linting/tests are failing on self enrollment stuff? Why was this commented out? Same for the TACheckinTimesResponse

IrisLiu-00 (changes requested):
Looking good so far, put in some questions and requested changes

It would be nice to test out the login endpoint using postman or something, if you guys havent already - I'll also do a bit of manual testing before I give final PR approval.

packages/server/src/login/course-section-mapping.entity.ts
brianyyang: The migration file has been committed 🎉

packages/server/src/login/login-course.service.ts
brianyyang: That makes sense, will attempt this now

packages/api-client/index.ts
brianyyang: I think it was a linting issue once we commented out the previous registration form

packages/server/src/login/login-course.service.ts
dfarooq610: For some more information: the Admin data will probably be formatted something like `Summer_Full 21`. 

Internally, we use the full year 2021 to prevent Y21K, so you make need to do a small bit of logic on the year there to make sure you query the right semester.

IrisLiu-00: ^ admin semester string format is in that notion doc outlining the admin post request

brianyyang: tyty

IrisLiu-00 (approved):
LGTM! I think you should be good to merge once you get another approval.

packages/server/src/login/login-course.service.ts
```diff
@@ -31,38 +31,36 @@ export class LoginCourseService {
...
+            this.convertKhouryRole(c.role),
           );
-          userCourses.push(taCourse);
+          userCourses.push(userCourse);
+        }
+      } else {
+        // c is a KhouryProfCourse
+        // only need to inspect one of the CRNs from the list
+        // b/c they should all map to the same section group
+        if (c.crns.length !== 0) {
```
dfarooq610, line 49: Can we throw a 400 otherwise? So admin can get an error message if they give us bad data

packages/server/src/login/login-course.service.ts
```diff
@@ -134,4 +136,37 @@ export class LoginCourseService {
...
+    let season = khourySeason;
+    switch (
+      khourySeason // summer sems are the only ones that are diff
+    ) {
+      case 'Summer I':
+        season = 'Summer_1';
+        break;
+      case 'Summer II':
+        season = 'Summer_2';
+        break;
```
dfarooq610, line 168: @jtavera235 just confirming one final time, the summer courses are formatted using Roman numerals like this, right?

dfarooq610: Yea just wanted to get your eyes on this parsing so we are 100% on the same page

packages/server/src/login/login-course.service.ts
IrisLiu-00: In the interest of getting things moving, I think we can also come back to this when everything else is implemented.

packages/server/src/login/login-course.service.ts
```diff
@@ -31,38 +31,36 @@ export class LoginCourseService {
...
+            course.id,
+            this.convertKhouryRole(c.role),
           );
-          userCourses.push(taCourse);
+          userCourses.push(userCourse);
+        }
+      } else {
+        // c is a KhouryProfCourse
+        // only need to inspect one of the CRNs from the list
+        // b/c they should all map to the same section group
```
jtavera235, line 48: This assumption is not necessarily true. An instructor can be part of multiple section groups as they can be teaching multiple different courses where each course belongs to a different section group so we would have to iterate through all the courses

packages/server/src/login/login-course.service.ts
jtavera235: So this is on me but turns out the format of the semester is going to be different than originally planned. The admin page is going to return the semester code that appears on Banner. This is a 6 digit string where the first 4 digits are the year and the last 2 digits represent the semester with the semester mappings being:
10 - Fall
30 - Spring
40 _ sum1
50 - SumF
60 - Sum2
However when it is the Fall semester, it maps to the academic year. As an example:
202110 is Fall of 2020 - The academic year is 2021 but the current year of the fall semster is 2020.
202130 is Spring of 2021, etc. The weird rule only applies to Fall Semester

jtavera235 (changes requested):
Really Great job! I left a comment with new changes in the semester format that I just learned about and a comment regarding an assumption that all courses will belong to the same section group. Other than that everything looks good

packages/server/src/login/login-course.service.ts
dfarooq610: This is analyzing a single section group when iterating through all the prof's courses. So when trying to check if one of the professor's section groups is registered for the app, I think it should be fine to check one CRN inside of that. Either all or zero CRNs will be registered inside a Khoury professor course

jtavera235: Ohhh my bad yep all good!

jtavera235 (approved):

packages/server/src/login/login-course.service.ts
```diff
@@ -142,31 +142,25 @@ export class LoginCourseService {
...
-      case 'Summer I':
-        season = 'Summer_1';
-        break;
-      case 'Summer II':
-        season = 'Summer_2';
-        break;
+    let year = Number(khourySemester.slice(0, 4));
+    const season = courseSeasonMap[khourySemester.slice(-2)];
+    // edge case for Fall semester
+    if (season == 'Fall') {
```
IrisLiu-00, line 161: while you're fixing tests - do triple equals, please!

dfarooq610 (approved):

PR #777, opened 2021-11-24
Start teams call if room name begins with "Online"

### LINKED ISSUE #763: [Queue Refactor] Only open Teams call if the Queue.room starts with "Online"
Sorry for the dev thrash on this -_-

Previously we were opening Teams if the room === "Online", but to facilitate multiple profs having their own Online queues, we're going to open Teams if the room starts with "Online".

VKong6019:
# Description

Small refactor on how teams would be opened. Previously only checked if the room name is just "Online", but now we check if the name begins with "Online".

Closes #763

## Type of change

# How Has This Been Tested?

- [x] Manually checking if teams opened

# Checklist:

### DIFF
--- a/packages/app/components/Queue/Student/StudentBanner.tsx
+++ b/packages/app/components/Queue/Student/StudentBanner.tsx
@@ -54,7 +54,7 @@ export default function StudentBanner({
   leaveQueue,
 }: StudentBannerProps): ReactElement {
   const { studentQuestion, studentQuestionIndex } = useStudentQuestion(queueId);
-  const isQueueOnline = useQueue(queueId).queue?.room === "Online";
+  const isQueueOnline = useQueue(queueId).queue?.room.startsWith("Online");
 
   switch (studentQuestion?.status) {
     case "Drafting":
--- a/packages/app/hooks/useTeams.tsx
+++ b/packages/app/hooks/useTeams.tsx
@@ -1,0 +1,15 @@
+import { useQueue } from "./useQueue";
+
+export function useTeams(queueId: number, users: string, message: string) {
+  const isQueueOnline = useQueue(queueId).queue?.room.startsWith("Online");
+
+  function openTeams() {
+    if (isQueueOnline) {
+      window.open(
+        `https://teams.microsoft.com/l/chat/0/0?users=${users}&message=${message}`
+      );
+    }
+  }
+
+  return openTeams;
+}
--- a/packages/app/hooks/useTeams.tsx
+++ b/packages/app/hooks/useTeams.tsx
@@ -1,9 +1,13 @@
 import { useQueue } from "./useQueue";
 
-export function useTeams(queueId: number, users: string, message: string) {
+export function useTeams(
+  queueId: number,
+  users: string,
+  message: string
+): () => void {
   const isQueueOnline = useQueue(queueId).queue?.room.startsWith("Online");
 
-  function openTeams() {
+  function openTeams(): void {
     if (isQueueOnline) {
       window.open(
         `https://teams.microsoft.com/l/chat/0/0?users=${users}&message=${message}`
--- a/up
+++ b/up
@@ -1,0 +1,2 @@
+yarn run v1.22.11
+info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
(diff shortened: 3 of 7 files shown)


up
```diff
@@ -0,0 +1,2 @@
+yarn run v1.22.11
```
VKong6019, line 1: what is this file lol

packages/app/components/Queue/Student/StudentBanner.tsx
```diff
@@ -54,7 +54,7 @@ export default function StudentBanner({
   leaveQueue,
 }: StudentBannerProps): ReactElement {
   const { studentQuestion, studentQuestionIndex } = useStudentQuestion(queueId);
-  const isQueueOnline = useQueue(queueId).queue?.room === "Online";
+  const isQueueOnline = useQueue(queueId).queue?.room.startsWith("Online");
```
IrisLiu-00, line 57: I did a full text search of the codebase for usages of the `teams.microsoft` link and it looks like there are a few places where we forgot to put in the `isOnline` check in the first place 💀 🤦 
* [when we open a call for question grouping](https://github.com/sandboxnu/office-hours/blob/58dbb1e1b92a58453add7930d1122596c4030364/packages/app/components/Queue/TA/QuestionGrouping/AllQuestionsChecklist.tsx#L106)
* [the 'help next' button on the queue page](https://github.com/sandboxnu/office-hours/blob/954592349d63c09e1a7696e8b4883434b749b254/packages/app/components/Queue/TA/TAQueue.tsx#L108)
   * It looks like I also forgot to swap the default teams message here to use the `useDefaultMessage` hook, if you could also do that it would be awesome!!

It kind of feels like we've run into issues with forgetting to update all the places that open a teams call several times now, so I'm wondering if it would be worth making some sort of util to abstract that logic.. you could maybe do a function that given some information like the users and default msg, creates the teams link and calls `window.open` with it? Or idk maybe a custom hook that returns a function to open the window?

IrisLiu-00 (changes requested):

packages/app/components/Queue/Student/StudentBanner.tsx
VKong6019: Aye aye captain

VKong6019: oh wait i have a question: am i applying the `isOnline` check in those places you mentioned in the `disabled` field or conditionally opening the window with an `if`?

for example in `AllQuestionsChecklist.tsx`, this snippet is in `BannerPrimaryButton`:
```
      window.open(
        `https://teams.microsoft.com/l/chat/0/0?users=${usersInLink}&message=${defaultMessage}`
      );
    }}
    disabled={!canHelp || checkedQuestions.size === 0}
```

cause if it's the former, i think it would be hard to abstract. also, the two files that i already refactored use both ways repsectively

IrisLiu-00: Hm I'm not sure I 100% understand what you're saying, but the idea is to just conditionally do `window.open` based on if the queue is online. The `disabled` is separate from that - eg. you might not be able to help someone because you're already helping someone else, which is unrelated to the queue being online. 

Conditionally calling window.open also may or may not be in that helper you abstract out, depending on what you decide to go with 🤷

VKong6019: I understand the assignment now; thanks chief

packages/app/hooks/useTeams.tsx
```diff
@@ -0,0 +1,19 @@
+import { useQueue } from "./useQueue";
+
+export function useTeams(
```
VKong6019, line 3: have mercy im a backend trying to do frontend 😓

packages/app/components/Queue/Student/StudentBanner.tsx
VKong6019, line 57: did not abstract this file cause i think `isQueueOnline` is being used to show/hide some component

VKong6019:
someone gib me vercel access D:

VKong6019:
cypress failing to run from yarn build
image

up
dfarooq610: Hmm try deleting it? Maybe that will get things to run?

dfarooq610 (approved):
LGTM

IrisLiu-00 (approved):
LGTM - Didn't manual test the new stuff but I'm gonna assume you tested it well. Go ahead and merge once the build is fixed.

PR #784, opened 2021-12-03
fix: load semester during calendar parsing

dfarooq610:
# Description
`course.semester` was nonexistent when parsing the ical, so we loaded the semester in. Also, fixed up the seeds controller for the SemesterModel so we dont have infinite semesters locally. 

## Type of change

# How Has This Been Tested?
Imported my own Ical File (not in this PR bc it's a secret), and manually hit the endpoint to update the calendar. We replicated the error, and after making these changes, we could successfully parse the calendar.

### DIFF
--- a/packages/server/src/course/ical.service.ts
+++ b/packages/server/src/course/ical.service.ts
@@ -191,9 +191,10 @@ export class IcalService {
       where: { courseId: course.id, room: 'Online' },
     });
     const semester =
+      course.semester ||
       (await SemesterModel.findOne({
         where: { id: course.semesterId },
-      })) || course.semester;
+      }));
 
     if (!queue) {
       queue = await QueueModel.create({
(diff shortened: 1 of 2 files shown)


ashdawngary (approved):
looks good

packages/server/src/course/ical.service.ts
```diff
@@ -190,6 +190,10 @@ export class IcalService {
     let queue = await QueueModel.findOne({
       where: { courseId: course.id, room: 'Online' },
     });
+    const semester =
+      (await SemesterModel.findOne({
+        where: { id: course.semesterId },
+      })) || course.semester;
```
IrisLiu-00, line 196: is `course.semester` supposed to go first in the `OR` here? this way we dont run the db query if we already have the sem.

IrisLiu-00 (approved):
Thanks for getting this, guys!! 🙏 🙏

packages/server/src/course/ical.service.ts
ashdawngary: yes i can fix that -- in prod course.semester is null its just the factories that dont give a semesterid but give a semestermodel object

PR #785, opened 2021-12-04
Database models for login flow

IrisLiu-00:
# Description
Updates/creates the database models outlined in our [problem doc](https://www.notion.so/sandboxnu/New-Login-Flow-42e109bacd54478a9c5e97256cd1b05b), with the exception of the semester mapping model changes which are implemented in #773 

This also depends on #773 because I need the Khoury Admin types from it. 

TODO: 
- [x] uncomment sectionGroups column
- [x] create migrations

## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/packages/server/src/course/course.entity.ts
+++ b/packages/server/src/course/course.entity.ts
@@ -30,7 +30,7 @@ export class CourseModel extends BaseEntity {
   @Column('text')
   name: string; // display name entered by Prof
 
-  @Column('text')
+  @Column('text', { nullable: true }) // nullable for backwards compatibility
   sectionGroupName: string; // from admin
 
   @Column('text', { nullable: true })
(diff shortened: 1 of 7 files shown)


packages/server/src/course/course.entity.ts
```diff
@@ -28,7 +28,10 @@ export class CourseModel extends BaseEntity {
   queues: QueueModel[];
 
   @Column('text')
-  name: string;
+  name: string; // display name entered by Prof
+
+  @Column('text', { nullable: true }) // nullable for backwards compatibility
+  sectionGroupName: string; // from admin
```
dfarooq610, line 34: Should we write a backfill to make old courses have the same name and question group name?

dfarooq610 (approved):

dfarooq610:
LGTM otherwise

packages/server/src/course/course.entity.ts
IrisLiu-00: Hmm I can log a ticket to come back to this once the other login stuff is done? That's probably nice to have but idt we actually will use the sectionGroupName for anything immediately so it might be fine if it's just null :P

#793

PR #790, opened 2021-12-17
Queue Create and check-in modals

### LINKED ISSUE #771: [ Queue Refactor ] -- Implement the new queue creation/check-in flow
Depends on #717
** we may want to decide whether or not to break up this issue as it has a couple diff BE/FE portions

* Implement Megan's figma designs for the queue creation modal + checking into an existing queue modal
* This should probably also involve deleting old code/components that will now no longer be used. Not sure how much of it there is, but stuff like the button TAs use for check in (since TAs and profs now have basically the same check in)
* Update queue creation to set `isProfQueue` based on what was entered in the FE form
* Prevent people from creating queues with duplicate room. In the old modal, if you enter a duplicate name the app just checks you into that preexisting queue. Since queue creation now involves editing some settings we probably want to disable that. 
    * Serve up an error from the backend endpoint if the room (name) is a duplicate
    * Show the error on the frontend

ashdawngary:
# Description

Made to match the figma designs. Note: isOnline switch is not in use right now since there is no analagous field for queues.  Also created a new endpoint (`POST :id/generate_queue/:room`) in courses to create queues with a predefined notes, isProfessorQueue, name, and room.  Wrote tests for the new endpoint, and updated cypress tests to work with new interface.  Allow TA's is perma-on for TA's creating queues and toggleable for professors.

(need to fix some FE alignment things but functional and tested)

Demos below:

qcc
(check into a queue)

qcr
(create a queue from scratch)

queue-alreadyexists
(create a joinable, but already existing queue)

queue-createprofessorexists
(create a unjoin-able, but already exists professor queue)

Closes #771 

## Type of change

# How Has This Been Tested?

- [x] Cypress tests in `queue_create_inperson_or_online.js` tests the flow of the modals.
- [x] new endpoint tested in `course.integration.ts`

# Checklist:

### DIFF
--- a/packages/app/components/Today/QueueCheckInButton.tsx
+++ b/packages/app/components/Today/QueueCheckInButton.tsx
@@ -1,41 +1,43 @@
 import { API } from "@koh/api-client";
 import { QueuePartial, Role } from "@koh/common";
-import { Form, Input, message, Modal, Radio } from "antd";
+import { Form, message, Modal, Select } from "antd";
 import { useRouter } from "next/router";
-import { Row } from "antd";
 import React, { ReactElement, useState } from "react";
-import styled from "styled-components";
 import { useCourse } from "../../hooks/useCourse";
 import { useProfile } from "../../hooks/useProfile";
 import { useRoleInCourse } from "../../hooks/useRoleInCourse";
 import TACheckinButton, { CheckinButton } from "./TACheckinButton";
 
-const ModalRadio = styled(Radio)`
-  display: block;
-  height: 30px;
-`;
-
 export default function TodayPageCheckinButton(): ReactElement {
   const profile = useProfile();
   const [modalVisible, setModalVisible] = useState(false);
-  const [queueToCheckInto, setQueueToCheckInto] = useState(0);
+  const [queueToCheckInto, setQueueToCheckInto] = useState(-1);
   const router = useRouter();
   const { cid } = router.query;
   const { course } = useCourse(Number(cid));
   const role = useRoleInCourse(Number(cid));
   const [form] = Form.useForm();
+  const { Option } = Select;
   const queueCheckedIn = course?.queues.find((queue) =>
     queue.staffList.find((staff) => staff.id === profile?.id)
   );
+
+  function onQueueUpdate(queueIx: number) {
+    setQueueToCheckInto(queueIx);
+  }
 
   return (
     <>
       {modalVisible && (
         <Modal
-          title="Check-In To Office Hours"
+          title="Check into an existing queue"
           visible={modalVisible}
-          onCancel={() => setModalVisible(false)}
+          onCancel={() => {
+            setModalVisible(false);
+            setQueueToCheckInto(-1);
+          }}
           okText="Check In"
+          okButtonProps={{ disabled: queueToCheckInto < 0 }}
           onOk={async () => {
             const value = await form.validateFields();
             let redirectID: QueuePartial;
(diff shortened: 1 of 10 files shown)


packages/app/components/Today/QueueCreateButton.tsx
```diff
@@ -0,0 +1,144 @@
...
+          title={`Create a new queue`}
+          visible={modalVisible}
+          onCancel={() => {
+            setModalVisible(false);
+          }}
+          okText="Create"
+          onOk={submitMakeQueue}
+        >
+          <Form form={form}>
+            <Row>
```
nicoledanuwidjaja, line 65: Gonna be *that person* but just a nit: a [Checkbox](https://ant.design/components/checkbox/) would probably look a bit less awkward?

packages/app/components/Today/QueueCheckInButton.tsx
```diff
@@ -60,45 +54,30 @@ export default function TodayPageCheckinButton(): ReactElement {
             }
           }}
         >
-          <h3>Which queue would you like to check into?</h3>
-          <Radio.Group
-            value={queueToCheckInto}
-            onChange={(e) => setQueueToCheckInto(e.target.value)}
+          <h3>Begin searching for a room</h3>
+          <Select
```
nicoledanuwidjaja, line 58: I couldn't test it on staging (it ded) but wondering if the dropdown would overflow the  modal or just like hide

nicoledanuwidjaja (commented):
👍

packages/app/components/Today/QueueCheckInButton.tsx
ashdawngary: it has a max size and it is scrollable within that window

packages/app/components/Today/QueueCreateButton.tsx
```diff
@@ -0,0 +1,146 @@
...
+                rules={[
+                  {
+                    required: true,
+                    message: "Please give this room a name.",
+                  },
+                ]}
+                initialValue={
+                  role === Role.TA
+                    ? ""
+                    : `Professor ${profile.lastName}'s Office Hours`
```
IrisLiu-00, line 113: I think we probably don't need this initialValue for professors, since the idea is the make queues based off of room names anyways

IrisLiu-00 (changes requested):
Looking good so far! The spacing on that queue creation form could be adjusted like Nicky mentioned, and I left a few other requests as well --

packages/app/components/Today/QueueCreateButton.tsx
```diff
@@ -51,6 +52,18 @@ export default function TodayPageCreateButton(): ReactElement {
...
+  function onIsOnlineUpdate(e) {
+    setRoomEditDisabled(e.target.checked);
+    if (e.target.checked) {
+      form.setFieldsValue({
+        officeHourName:
+          role === Role.TA
+            ? "Online"
+            : `Online - Professor ${profile.lastName}'s Office Hours`,
+      });
+    }
```
IrisLiu-00, line 64: If the checkbox becomes unchecked, let's set the field value back to empty.

packages/app/components/Today/QueueCreateButton.tsx
```diff
@@ -51,6 +52,18 @@ export default function TodayPageCreateButton(): ReactElement {
...
   }
 
+  function onIsOnlineUpdate(e) {
+    setRoomEditDisabled(e.target.checked);
+    if (e.target.checked) {
+      form.setFieldsValue({
+        officeHourName:
+          role === Role.TA
+            ? "Online"
+            : `Online - Professor ${profile.lastName}'s Office Hours`,
```
IrisLiu-00, line 62: This logic actually needs to be a bit more complicated - we only want to say `Professor X's office hours` if `allow TAs` is unchecked, because that toggle is what makes this queue a Prof queue. so you'll probably need another handler for the `allow TAs` checkbox which:
* Does nothing if we are not online
* sets or removes `Professor X's office hours` if we are online

Also while you're at it you might as well make these arrow functions too :P

IrisLiu-00 (commented):
Thanks for making changes! the room name autopopulation needs a bit more tweaking though - let me know if what I wrote doesn't make sense. 

The two checkboxes might also look a bit better if spaced out or broken into two rows, but I can probably do that if you're not as into styling shit lol

packages/app/components/Today/QueueCreateButton.tsx
IrisLiu-00: TBH it might just be easier to set the `value` prop on the room input if we are `Online`, instead of fiddling with the fields through a handler? Then this handler would only set editability of that field and clear the value when its unchecked? I think that would work but not sure

IrisLiu-00 (approved):

PR #794, opened 2021-12-23
Backend endpoint for handling pending courses for profs

### LINKED ISSUE #780: [ Login Flow ] -- Professor User Course Registration Backend
Update the login flow to incorporate the new professor export 
- [x] Update the data types for the `courses` field in the Khoury Data Params (in the notion doc)
- [x] Update the POSTs to /khoury_login endpoint to do the following if the user is a professor: 
* If we have not seen the semester in the body of the course, create a new Semester Model
* Save the whole list of section groups to ProfessorSectionGroupsModel 
 * If the professor has not logged in this semester (using LastRegistrationModel), add pendingCourses to GetProfileResponse, which has any section groups from ProfessorSectionGroupModel that do not already exist as CourseModels.

More info can be found here 
https://www.notion.so/sandboxnu/New-Login-Flow-42e109bacd54478a9c5e97256cd1b05b

brianyyang:
# Description

Updates the GET/profile endpoint to return pending courses for professors if there are any. This will eventually be used in the frontend to redirect to course registration page on login.

There are also some changes to the POST/khoury_login endpoint:
* Save all the professor's sectionGroup data to DB 
* Creates the SemesterModel if none found

Closes #780 (this issue overlapped with #743 but there were some things that weren't covered)

## Type of change

# How Has This Been Tested?

We added integration and unit tests in `profile.integration.ts` and `profile.service.spec.ts` respectively.

# Checklist:

### DIFF
--- a/packages/server/src/login/login-course.service.ts
+++ b/packages/server/src/login/login-course.service.ts
@@ -5,6 +5,7 @@ import { CourseSectionMappingModel } from 'login/course-section-mapping.entity';
 import { UserCourseModel } from 'profile/user-course.entity';
 import { UserModel } from 'profile/user.entity';
 import { Connection } from 'typeorm';
+import { ProfSectionGroupsModel } from './prof-section-groups.entity';
 
 @Injectable()
 export class LoginCourseService {
@@ -77,6 +78,14 @@ export class LoginCourseService {
           userCourses.push(previousCourse);
         }
       }
+    }
+
+    // If Prof, save the JSON data
+    if (info.courses[0] && !isKhouryCourse(info.courses[0])) {
+      await ProfSectionGroupsModel.create({
+        profId: user.id,
+        sectionGroups: info.courses,
+      }).save();
     }
 
     user.courses = userCourses;
--- a/packages/server/src/login/login-course.service.ts
+++ b/packages/server/src/login/login-course.service.ts
@@ -4,6 +4,7 @@ import { CourseModel } from 'course/course.entity';
 import { CourseSectionMappingModel } from 'login/course-section-mapping.entity';
 import { UserCourseModel } from 'profile/user-course.entity';
 import { UserModel } from 'profile/user.entity';
+import { SemesterModel } from 'semester/semester.entity';
 import { Connection } from 'typeorm';
 import { ProfSectionGroupsModel } from './prof-section-groups.entity';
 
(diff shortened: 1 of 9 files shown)


packages/server/src/login/login-course.service.ts
```diff
@@ -88,14 +98,13 @@ export class LoginCourseService {
     courseCRN: number,
     semester: string, // 6-digit semester code
   ): Promise<CourseModel> {
-    const { season, year } = this.parseKhourySemester(semester);
+    const semModel = await this.getSemester(semester);
     const courseSectionModel =
       await CourseSectionMappingModel.createQueryBuilder('section_mapping')
         .leftJoinAndSelect('section_mapping.course', 'course')
```
dfarooq610, line 104: wow does the 'course' alias work? that's awesome 🤯

packages/server/src/profile/profile.service.ts
```diff
@@ -0,0 +1,49 @@
...
+        profId: userId,
+      },
+    });
+    const lastRegistered = await LastRegistrationModel.findOne({
+      where: {
+        profId: userId,
+      },
+    });
+
+    if (!profCourses) return; // not a professor
```
dfarooq610, line 27: should the pending courses not exist in the profile payload at all if the professor has already registered? Or should we use the empty list? 

I think this would affect how we want to redirect

packages/server/src/profile/profile.service.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+      const student = await UserFactory.create();
+      const resp = await service.getPendingCourses(student.id);
+      expect(resp).toBeUndefined();
+    });
+
+    it('returns pending courses (sans already registered courses) for a prof', async () => {
+      const resp = await service.getPendingCourses(prof1.id);
+      expect(resp).toEqual([prof1KhouryCourses[1]]);
+    });
+
```
dfarooq610, line 88: smol but can you add a quick test for a professor that has registered in a previous semester and now has new courses

packages/server/src/profile/profile.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (lastRegistered?.lastRegisteredSemester !== profCourseSem) {
+      // current semester does not match last time prof registered, there may be pending courses
+      for (const c of profCourses.sectionGroups) {
+        if (c.crns.length !== 0) {
+          const courseCRN = c.crns[0];
+          const profCourse = await this.loginCourseService.courseCRNToCourse(
+            courseCRN,
+            c.semester,
+          );
+          if (!profCourse) pendingCourses.push(c);
```
dfarooq610, line 43: inline if statements give me the heebie jeebies but its alright

dfarooq610 (changes requested):
this is nice! I left a small comment about the pendingCourses payload for profs and asked for one more test

packages/server/src/profile/profile.service.ts
IrisLiu-00: Ermm yeah we were just gonna check for both the undefined and empty array case on the FE haha.. partially bc its possible that the prof hasn't already registered for this sem, but all their courses have actually been registered already by other profs (in which case the logic ends up generating an empty array as well)

We can add some more logic here to ensure the function only ever returns either undefined or empty array if you feel strongly about it, though.

dfarooq610: Ahh ok yea that would work -- personally I think it's better to have the pending courses be populated or nonexistent (one less check on the frontend lol). However it's fine either way

dfarooq610 (approved):

Issue #798, opened 2021-12-24
Don't allow users to go to the page for a disabled queue


IrisLiu-00:
Currently, users can still navigate to the url of a disabled queue and check in or submit questions. We should prevent this. 

There are a couple options here (possibly more):
* Serve up some error from the backend when we try to GET that queue
* Have the frontend redirect away from the page
* Have the frontend show some error message and disable interaction with the queue

PR #800, opened 2021-12-27
update QueueModel queries for isDisabled

### LINKED ISSUE #797: Testing for Disabling queues feature
Regression test and investigate the codebase for any inconsistencies when fetching queues that are disabled. Either create issues or submit PRs to fix them.

ashdawngary:
# Description

updated all lookups that use the `room` field as a a key to look for non-disabled queues.  Updated course controller and ical service. Grepped across server/ to look for instances of `QueueModel.find`.  Also added indicators on disabled queues to tell the user that is queue disabled.  

Checking into a disabled queue yields a 404 on POST, added a message that propagates the error.  Also disabled the disable queue button, since the queue is disabled.

disable-safeguards

Closes #797 and #798

## Type of change

# How Has This Been Tested?

- [ ] Tests added to endpoint to test for location key collision.

# Checklist:

### DIFF
--- a/packages/app/components/Queue/QueueListSharedComponents.tsx
+++ b/packages/app/components/Queue/QueueListSharedComponents.tsx
@@ -143,7 +143,9 @@ export function QueueInfoColumn({
   return (
     <InfoColumnContainer>
       <QueueRoomGroup>
-        <QueueTitle data-cy="room-title">{queue?.room}</QueueTitle>
+        <QueueTitle data-cy="room-title">
+          {queue?.room} {queue?.isDisabled && <b>(disabled)</b>}
+        </QueueTitle>
         {!queue.allowQuestions && (
           <Tooltip title="This queue is no longer accepting questions">
             <StopOutlined
@@ -189,8 +191,9 @@ export function QueueInfoColumn({
         <DisableQueueButton
           onClick={confirmDisable}
           data-cy="queue-disable-button"
+          disabled={queue?.isDisabled}
         >
-          Disable Queue
+          {queue?.isDisabled ? `Queue disabled` : `Disable Queue`}
         </DisableQueueButton>
       )}
     </InfoColumnContainer>
--- a/packages/app/components/Queue/Student/StudentQueue.tsx
+++ b/packages/app/components/Queue/Student/StudentQueue.tsx
@@ -267,7 +267,7 @@ export default function StudentQueue({ qid }: StudentQueueProps): ReactElement {
   );
 
   if (queue && questions) {
-    if (!queue.isOpen) {
+    if (!queue.isOpen || queue.isDisabled) {
       return <h1 style={{ marginTop: "50px" }}>The Queue is Closed!</h1>;
     }
     return (
--- a/packages/app/components/Queue/Student/StudentQueue.tsx
+++ b/packages/app/components/Queue/Student/StudentQueue.tsx
@@ -299,7 +299,7 @@ export default function StudentQueue({ qid }: StudentQueueProps): ReactElement {
                 >
                   <JoinButton
                     type="primary"
-                    disabled={!queue?.allowQuestions}
+                    disabled={!queue?.allowQuestions || queue?.isDisabled}
                     data-cy="join-queue-button"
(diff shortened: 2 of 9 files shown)


packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -189,8 +191,9 @@ export function QueueInfoColumn({
         <DisableQueueButton
           onClick={confirmDisable}
           data-cy="queue-disable-button"
+          disabled={queue?.isDisabled}
```
IrisLiu-00, line 194: It might be worth it to also disable the Check In button (for TAs) and the Join Queue button (for students) if the queue is disabled - it looks like those buttons are passed in via `TAQueue.tsx` and `StudentQueue.tsx`. We could put a tooltip on the button explaining that you can't check into disabled queues as well.

packages/server/src/course/course.controller.ts
```diff
@@ -370,6 +371,7 @@ export class CourseController {
       {
         room,
         courseId,
+        isDisabled: false,
```
IrisLiu-00, line 374: I wonder if it's also worth adding a `disabled` check to `QueueModel.checkIsOpen()` - it looks like the `createQuestion` endpoint doesn't explicitly check that the question is being created on a disabled queue, but it does do `checkIsOpen`. `checkIsOpen` just checks if the queue has any staff, and theoretically a disabled queue should never have any staff - but it might be nice to also make sure that it's also not disabled to be safe.

IrisLiu-00 (commented):
This is great, thanks for going through and testing stuff! I've just got a few small suggestions this time -

packages/app/components/Queue/QueueListSharedComponents.tsx
ashdawngary: apparently tooltips dont show up on disabled elements :(

dfarooq610: We have something like this somewhere I think? Like if you try to help a student when you aren't checked in

packages/app/components/Queue/Student/StudentQueue.tsx
```diff
@@ -267,7 +267,7 @@ export default function StudentQueue({ qid }: StudentQueueProps): ReactElement {
   );
 
   if (queue && questions) {
-    if (!queue.isOpen) {
+    if (!queue.isOpen || queue.isDisabled) {
```
IrisLiu-00, line 270: if we updated the checkIsOpen function to set isOpen based on being disabled as well, do we need to also check queue.isDisabled?

packages/app/components/Queue/QueueListSharedComponents.tsx
IrisLiu-00: Yeah [heres an example](https://github.com/sandboxnu/office-hours/pull/800/files/1cadd2f188d2bdb387dbf3d68be3f42a502638fa..8ec16ff01a89284e40000c993fb96dc39aeb6a73#diff-a76a51c3539f561d5b49ce8fdc4ea95961d162134f96dad909f3653acedab916R134) of a conditional tooltip on a disabled button

packages/app/components/Queue/TA/TAQueue.tsx
```diff
@@ -142,14 +142,16 @@ export default function TAQueue({ qid, courseId }: TAQueueProps): ReactElement {
...
                 <div style={{ marginBottom: "12px" }}>
                   <TACheckinButton
                     courseId={courseId}
                     room={queue?.room}
                     disabled={
                       staffCheckedIntoAnotherQueue ||
                       isHelping ||
-                      (queue.isProfessorQueue && role !== Role.PROFESSOR)
+                      (queue.isProfessorQueue && role !== Role.PROFESSOR) ||
+                      queue.isDisabled
```
IrisLiu-00, line 154: Could you also do this disabling for the `Join` button that students get (its in `StudentQueue.tsx`) 🙏

IrisLiu-00 (approved):
LGTM

PR #801, opened 2021-12-30
Register course endpoint

nicoledanuwidjaja:
# Description
As part of the new login flow, this new endpoint uses a new array of `RegisterCourseParams` which allows professors to register multiple `SectionGroups` at once. (A `SectionGroup` represents a course with multiple section offerings, denoted by CRN)

- Refactors the `CourseModel` object to have a `sectionGroupName` field
- Generates `CourseSectionMappings` for each item in `sectionGroup.crns`
- Creates a `UserCourse` where professor is assigned as a user of the created course
- Updates professor's last registered semester to match the course's semester

Closes #781.

## Type of change
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?
New course integration test for `register_courses` endpoint.

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code where needed 
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #807, opened 2022-01-03
[Login Flow] Register Course Endpoint

### LINKED ISSUE #781: [ Course Onboarding ] -- Create endpoint to register new courses
Create POST /courses/registerCourses (scoped to a professor role). 

- Post body is formatted like this
    ```jsx
    [
    	{
    		"name": string,  <---- Section Group Name
                    "displayName": string,  <--- user friendly display name entered by Prof
    		"crns": [int],   <---- List of CRN's in the section group
    		"semester": string,
    		"iCalURL": string,
    		"coordinatorEmail": string,
    		"timezone": string,  <--- derived from the Campus field on the form  
    	},
    	...
    ]
    ```

More information found here https://www.notion.so/sandboxnu/New-Login-Flow-42e109bacd54478a9c5e97256cd1b05b

Objectives:
- Verify that the posted section groups are ones the professor has from khoury admin (check against `ProfessorSectionGroupsModel`)
- Create CourseModels for each section group (set to enabled)
- Create CourseSectionMappings with each CRN
- Create UserCourses between the Prof and each CourseModel
- Set `LastRegistrationModel.lastRegisteredSemester` to the semester of the posted courses

tiingweii-shii:
# Description
As part of the new login flow, this new endpoint uses a new array of `RegisterCourseParams` which allows professors to register multiple `SectionGroups` at once. (A `SectionGroup` represents a course with multiple section offerings, denoted by CRN)

- Refactors the `CourseModel` object to have a `sectionGroupName` field
- Generates `CourseSectionMappings` for each item in `sectionGroup.crns`
- Creates a `UserCourse` where professor is assigned as a user of the created course
- Updates professor's last registered semester to match the course's semester

Closes #781.
This PR was created because the [original one](https://github.com/sandboxnu/office-hours/pull/801) was closed by accident. 

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?
New course integration test for `register_courses` endpoint.

# Checklist:

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -808,6 +808,7 @@ export const ERROR_MESSAGES = {
       cannotCheckIntoMultipleQueues:
         "Cannot check into multiple queues at the same time",
     },
+    courseAlreadyRegistered: "The course was already registered",
     courseNotFound: "The course was not found",
     sectionGroupNotFound: "The section group was not found",
     courseOfficeHourError: "Unable to find a course's office hours",
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -696,9 +696,6 @@ export class RegisterCourseParams {
   @IsString()
   displayName!: string;
 
-  @IsArray()
-  crns!: number[];
-
   @IsString()
   semester!: string;
 
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -696,9 +696,6 @@ export class RegisterCourseParams {
   displayName!: string;
 
   @IsString()
-  semester!: string;
-
-  @IsString()
   iCalURL!: string;
 
   @IsString()
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -808,9 +808,9 @@ export const ERROR_MESSAGES = {
       cannotCheckIntoMultipleQueues:
         "Cannot check into multiple queues at the same time",
     },
-    courseAlreadyRegistered: "The course was already registered",
+    courseAlreadyRegistered: "One or more of the courses is already registered",
     courseNotFound: "The course was not found",
-    sectionGroupNotFound: "The section group was not found",
+    sectionGroupNotFound: "One or more of the section groups was not found",
     courseOfficeHourError: "Unable to find a course's office hours",
     courseHeatMapError: "Unable to get course's cached heatmap",
     courseModelError: "Course Model not found",
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -809,6 +809,7 @@ export const ERROR_MESSAGES = {
         "Cannot check into multiple queues at the same time",
     },
     courseNotFound: "The course was not found",
+    sectionGroupNotFound: "The section group was not found",
     courseOfficeHourError: "Unable to find a course's office hours",
     courseHeatMapError: "Unable to get course's cached heatmap",
     courseModelError: "Course Model not found",
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -680,20 +680,20 @@ export class GetAlertsResponse {
 }
 
 /**
- * Represents the parameters for register_courses endpoint.
- * @param name - The name of the section group.
- * @param displayName - user friendly display name entered by Prof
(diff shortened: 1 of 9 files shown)


packages/server/src/course/course.controller.ts
```diff
@@ -525,64 +522,23 @@ export class CourseController {
...
-        await CourseSectionMappingModel.create({
-          genericCourseName: body.name,
-          section,
-          courseId: course.id,
-        }).save();
-      });
+      this.courseService.registerCourses(body, userId);
     } catch (err) {
       console.error(err);
       throw new HttpException(
```
dfarooq610, line 579: You would want to throw the same err you caught from the service after logging it

packages/server/src/course/course.service.ts
```diff
@@ -101,4 +114,117 @@ export class CourseService {
...
+          profLastRegistered.lastRegisteredSemester = sectionGroup.semester;
+          await profLastRegistered.save();
+        } else {
+          profLastRegistered = await LastRegistrationModel.create({
+            profId: userId,
+            lastRegisteredSemester: sectionGroup.semester,
+          }).save();
+        }
+      } catch (err) {
+        console.error(err);
```
dfarooq610, line 222: I don't think you have to console.err here (since it will be caught in the controller and logged there)

packages/server/src/course/course.service.ts
```diff
@@ -101,4 +114,117 @@ export class CourseService {
...
+      });
+      if (!semester)
+        throw new BadRequestException(
+          ERROR_MESSAGES.courseController.noSemesterFound,
+        );
+
+      // checks that course hasn't already been created
+      let course = await CourseModel.findOne({
+        where: {
+          name: courseParams.name,
```
IrisLiu-00, line 154: I don't think matching by the name will work, since the name is just any string entered by a prof - you could have two profs enter different display names. Matching by the sectionGroupName should work, though.

packages/common/index.ts
```diff
@@ -805,7 +808,9 @@ export const ERROR_MESSAGES = {
       cannotCheckIntoMultipleQueues:
         "Cannot check into multiple queues at the same time",
     },
+    courseAlreadyRegistered: "The course was already registered",
```
IrisLiu-00, line 811: We made the front end to directly display these error messages in the case of an error, so let's make them a bit more user friendly and say something like, "One or more of the courses is already registered". Same goes for `sectionGroupNotFound`, let's make it account for the case where the prof is registering multiple courses as well.

packages/server/src/course/course.service.ts
IrisLiu-00: alternatively you could get rid of the console.err in the controller, since at that level you'll be catching your custom thrown HttpException -  I think it makes sense to log whatever low level typeORM error we get, then just throw the our custom HttpException.

IrisLiu-00 (changes requested):

packages/server/src/course/course.service.ts
dfarooq610: Section group name in the same semester

nicoledanuwidjaja: whoops, meant `sectionGroupName` here.

packages/server/src/course/course.controller.ts
```diff
@@ -525,64 +523,19 @@ export class CourseController {
...
-      // create CourseSectionMappings for each unique submitted section number
-      new Set(body.sections).forEach(async (section) => {
-        await CourseSectionMappingModel.create({
-          genericCourseName: body.name,
-          section,
-          courseId: course.id,
-        }).save();
-      });
+      await this.courseService.registerCourses(body, userId);
     } catch (err) {
```
IrisLiu-00, line 535: since you're just directly throwing this err you could just get rid of the whole try/catch

IrisLiu-00 (approved):
LGTM!

dfarooq610 (approved):
😎

PR #819, opened 2022-01-13
set lastregistered sem if no courses are registered

dfarooq610:
If profs decide not to register any new courses (this can be bc another professor has already registered on their behalf), we do not set the last registered semester for them. They'll see the apply page every time they log in. 

We set the last registered semester inside the for loop when analyzing each section group in the list. This PR pulls it out of the loop, which now means if the professor does not select any courses to register, we set the lastRegisteredSemester to the current semester's Khoury code (I made a new helper to find this). 

# Description

Please include a summary of the change and which issue is fixed.
Please also include relevant motivation and context. 
List any dependencies that are required for this change.

Closes # (issue number)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/server/src/course/course.service.ts
+++ b/packages/server/src/course/course.service.ts
@@ -127,6 +127,8 @@ export class CourseService {
       where: { profId: userId },
     });
 
+    let lastRegisteredSemester: string;
+
     // iterate over each section group registration
     for (const courseParams of body) {
       // finds professor's section group with matching name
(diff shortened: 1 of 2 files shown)


packages/server/src/course/course.service.ts
```diff
@@ -206,28 +208,69 @@ export class CourseService {
...
+
+    try {
+      // Update professor's last registered semester to semester model's current semester
+      let profLastRegistered: LastRegistrationModel;
+      profLastRegistered = await LastRegistrationModel.findOne({
+        where: { profId: userId },
+      });
+
+      // if the professor did not register anything, then find the current semester
+      if (!profLastRegistered) {
```
dfarooq610, line 222: Should be `!lastRegisteredSemester`

packages/server/src/course/course.service.ts
```diff
@@ -206,28 +208,69 @@ export class CourseService {
...
+      year -= 1;
+    }
+
+    return `${year}${courseSeasonMap[season]}`;
+  };
+
+  private monthToSeason(month: number): string {
+    if (1 <= month && month <= 4) {
+      return 'Spring';
+    } else if (5 <= month && month <= 6) {
```
nicoledanuwidjaja, line 268: Hmm. Would we have to look at the courses the professor is registered for and see if any of them fall under Summer_1/SummerFuLL? (Idk which one would be the default semester then)

packages/server/src/course/course.service.ts
```diff
@@ -206,28 +208,69 @@ export class CourseService {
...
+    };
+
+    if (season === 'Fall') {
+      year -= 1;
+    }
+
+    return `${year}${courseSeasonMap[season]}`;
+  };
+
+  private monthToSeason(month: number): string {
```
nicoledanuwidjaja, line 265: I think this *could* be problematic in cases where like the start of the semester is the beginning of a month, right? (e.g. May 1st is the start of summer 1, professor is lazily going through register flow in April) 

Could we do something where we match the current day by the lowest difference in number of days from the "start dates" of each semester? Is that too much?

ex: Today is April 26, 2022.
`Spring_startDate`: January 18, 2022
`Summer_1_startDate`: May 1, 2022
today - summer_1 = 4 days
today - spring = 100+ days
...therefore set today as `Summer_1` Semester.

nicoledanuwidjaja (commented):
Thoughts

packages/server/src/course/course.service.ts
nicoledanuwidjaja: I added a proof of concept in this [sandbox](https://replit.com/@nicoledanuwidja/FailingSlategrayNotifications#index.js)

dfarooq610: This is really good!! I think it will do the job.

PR #825, opened 2022-01-31
Handle Registering for Multiple Semesters (Summer Shenanigans)

### LINKED ISSUE #820: Handle Summer Course Registration
Our redirect to the course onboarding flow is based on if a professor's last registered semester is different than the semester of the courses we receive from Khoury admin. 

However the definition of that lastRegisteredSemester is a little cloudy if a professor is teaching both summer full and summer 1 courses. If they try to register for both, what lastRegisteredSemester should we use? What semester value do we use for deciding if there are any pending courses for registration? What happens if the courses in the Khoury Data Params have different semesters upon login?

Look into this problem and create tickets for whatever solution you come up with. Implement those changes in a PR.

This will require changes in:
* deciding if there are any pending courses, in ProfileService
* setting the lastRegisteredSemester in the course registration endpt

brianyyang:
# Description

Refactors our pending courses parsing to consider Summer 1 and Summer Full to be the same semester.

Previously, our redirect to the course onboarding flow is based on if a professor's last registered semester is different than the semester of the courses we receive from Khoury admin. This raises the issue of Summer 1/Summer Fall registration as if you register for both at the same time, it was unclear which to set `lastRegisteredSemester` to and how to handle that when later registering for Summer 2 classes.

Closes #820 

## Type of change

# How Has This Been Tested?

This has not been manually tested but we wrote two unit tests in `profile.service.spec.ts` for the case in which we want Summer 2 courses to be pushed to pending courses and `lastRegisteredSemester` is either Summer 1 or Summer 2. All the previous unit tests pass still as well so we probably didn't break anything.

# Checklist:

### DIFF
--- a/packages/server/src/profile/profile.service.ts
+++ b/packages/server/src/profile/profile.service.ts
@@ -26,6 +26,7 @@ export class ProfileService {
         profId: userId,
       },
     });
+    const lastRegisteredSemester = lastRegistered?.lastRegisteredSemester;
 
     if (!profCourses) return; // not a professor
 
@@ -35,10 +36,8 @@ export class ProfileService {
       // current semester does not match last time prof registered, there may be pending courses
       if (
         c.crns.length !== 0 &&
-        !this.isSameRegistrationTime(
-          c.semester,
-          lastRegistered?.lastRegisteredSemester,
-        )
+        (!lastRegisteredSemester || // lastRegistered doesnt exist if prof has never registered before
+          !this.isSameRegistrationTime(c.semester, lastRegisteredSemester))
       ) {
         const courseCRN = c.crns[0];
         const profCourse = await this.loginCourseService.courseCRNToCourse(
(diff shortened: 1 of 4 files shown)


packages/server/src/profile/profile.service.ts
```diff
@@ -23,27 +26,47 @@ export class ProfileService {
...
+
+  // hacky way to check if two semester codes from Khoury have the same registration window
+  private isSameRegistrationTime(semester1: string, semester2: string) {
+    const year1 = Number(semester1.slice(0, 4));
+    const year2 = Number(semester2.slice(0, 4));
+    const semesterCode1 = semester1.slice(-2);
+    const semesterCode2 = semester2.slice(-2);
+    // we want to treat Summer 1 and Summer Full as the same window
+    // for profs to register classes for both at the same time
+    if (
```
dfarooq610, line 62: Nit: just save the Summer1 and Summer full codes as constants so you don't have to access them twice

dfarooq610 (approved):
Seems pretty par for the course! Good job!

tiingweii-shii (approved):
Looks good :D

packages/server/src/profile/profile.service.ts
brianyyang: yes

PR #834, opened 2022-02-12
Make sure staffList is fetched when using queue.checkIsOpen()

IrisLiu-00:
# Description
Given that the checkIsOpen() method on the QueueModel now basically only depends on if there are people in the stafflist, it becomes pretty important that the stafflist relation is fetched when using this method - otherwise it returns the wrong result where it looks like the queue is not open because the stafflist array is undefined. 

I found this issue while working with Vera on #661 - the alerts service wasn't fetching the stafflist array ([now fixed](https://github.com/sandboxnu/office-hours/blob/vk-fix-fix-alerts/packages/server/src/alerts/alerts.service.ts#L29)), so alerts were getting improperly purged when the queue was actually still open. 

This PR just has the checkIsOpen() method throw an error if the staffList field doesn't exist - this isn't entirely ideal, as you won't realize you've forgotten the stafflist until the app actually runs. I tried seeing if there is a way to fetch relations on an object that you already got back from the ORM (so that checkIsOpen can just add the staffList to the queue if its not there), but I couldn't find anything. If anyone knows of something that would work lmk tho

As part of this I also looked for other usages of checkIsOpen() that were not fetching the stafflist - I didnt find anything, so the only alerts service bug should be fixed when that work gets merged in. 

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/packages/server/src/queue/queue.entity.ts
+++ b/packages/server/src/queue/queue.entity.ts
@@ -17,6 +17,8 @@ import { CourseModel } from '../course/course.entity';
 import { OfficeHourModel } from '../course/office-hour.entity';
 import { UserModel } from '../profile/user.entity';
 import { QuestionModel } from '../question/question.entity';
+import { HttpException, HttpStatus } from '@nestjs/common';
+import { ERROR_MESSAGES } from '@koh/common';
 
 interface TimeInterval {
   startTime: Date;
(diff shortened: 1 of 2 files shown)


packages/server/src/queue/queue.entity.ts
```diff
@@ -74,8 +76,14 @@ export class QueueModel extends BaseEntity {
   isOpen: boolean;
 
   async checkIsOpen(): Promise<boolean> {
-    this.isOpen =
-      this.staffList && this.staffList.length > 0 && !this.isDisabled;
+    if (!this.staffList) {
+      console.error(ERROR_MESSAGES.queueController.missingStaffList, this.id);
```
VKong6019, line 80: nit - we don't need to log this if we throw the exception right? we could include the `id` in the error message below

IrisLiu-00:
> would adding the `@eager` annotation on the `staffList` column help? or am i thinking of smth else

That's a good point.. we would just be always fetching staffList when we fetch a Queue 🤔 the downside is we may automatically fetch stafflist in some places where we don't need it. 

I combed through the code and there's a handful of places that fetch Queues without staffLists:
generateQueue()
groupQuestions()
queueRoleDecorater 
questionRoleGuard
queueRoleGuard

... and also a bunch of queries ran in tests.

PR #835, opened 2022-02-13
Backfill sectionGroupName

### LINKED ISSUE #793: [Course Onboarding] Backfill sectionGroupName
It might be good to write a backfill to set the sectionGroupName of older courses as the same as the course name. Then we can get rid of the nullable constraint on the column.

sumitde22:
# Description

Wrote a quick backfill to fill in the sectionGroupName column in course_model  for old cases where it's null

Allows that column to be marked non-nullable

Closes #793

## Type of change

- [] 
# How Has This Been Tested?

- [] Test A: Ran backfill locally through yarn cli and saw that dev table was updated correctly. 

# Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/server/src/backfill/backfill-section-group-name.ts
+++ b/packages/server/src/backfill/backfill-section-group-name.ts
@@ -1,0 +1,20 @@
+import { Injectable } from '@nestjs/common';
+import { CourseModel } from 'course/course.entity';
+import { Command } from 'nestjs-command';
+
+@Injectable()
+export class BackfillSectionGroupName {
+  @Command({
+    command: 'backfill:section-group-name',
+    describe:
+      'set the sectionGroupName of older courses as the same as the course name',
+    autoExit: true,
+  })
+  async fix(): Promise<void> {
+    await CourseModel.createQueryBuilder()
+      .update()
+      .set({ sectionGroupName: () => 'name' })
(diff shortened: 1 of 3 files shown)


packages/server/src/backfill/backfill-section-group-name.ts
```diff
@@ -0,0 +1,20 @@
...
+      'set the sectionGroupName of older courses as the same as the course name',
+    autoExit: true,
+  })
+  async fix(): Promise<void> {
+    await CourseModel.createQueryBuilder()
+      .update()
+      .set({ sectionGroupName: () => 'name' })
+      .where('sectionGroupName IS NULL')
+      .execute();
+  }
```
VKong6019, line 19: should we include a log message here to indicate the update occurred? i think logging how many courses were updated is good enough or smth to that effect

VKong6019 (commented):
i think other than smol comment. LGTM

packages/server/src/backfill/backfill-section-group-name.ts
sumitde22: Good idea, added

VKong6019 (approved):
Amazing

dfarooq610 (approved):
🎉

PR #859, opened 2022-03-10
Retrieve user info for professors endpoint

### LINKED ISSUE #843: Endpoint for professors to retrieve Students, TAs, and Profs
There should be a GET endpoint in a course that will return members of the course roster -- this will be used on the course roster page we are planning. For each user, we want:
1) their name
2) their email
3) their profile pic

Initial Thoughts: 
- This should be Role Guarded to Professors

- We probably want to paginate this, so we dont blow up memory by trying to fetch all users for a course at once (loading everyone in Fundies would break everything). 

- We might want a query param for the role we are looking for (TA | Student | Professor).

brianyyang:
# Description

This PR creates an endpoint for professors to retrieve list of users given a course id. The endpoint includes pagination, search by name substring matching, and search by role. The users are sorted alphabetically by first name when returned. Please review, this was disgusting. I hate typeORM.

Closes #843 

## Type of change

# How Has This Been Tested?

Manual Testing:
I used Postman for manual testing. You can send a GET request to `localhost:3000/api/v1/courses/3/get_user_info/1` with the DB and yarn dev running. This should get you all 5 users of the course. If you include `search` as a key, you can retrieve users based on name:

`search = e` should give you 4 users, all users minus Da-Jin.
`search = le` should give you 2 users, Alex and Stanley.
`search = Alex` should give you 1 user, Alex only.

Also searching should be case-insensitive so `alex` would also work fine.

To test pagination, you would have to either add more users (which I don't want to do) or go into the code and change PAGE_SIZE to a number that is less than 5 because there are only 5 users:

Unit Testing:
I wrote tests for general retrieval functionality, role flags, search terms, and pagination in a new spec file `course.service.spec.ts`

# Checklist:

### DIFF
--- a/packages/server/src/course/course.controller.ts
+++ b/packages/server/src/course/course.controller.ts
@@ -557,17 +557,33 @@ export class CourseController {
     }
   }
 
-  @Get(':id/get_user_info/:page/:search')
-  @UseGuards(JwtAuthGuard, CourseRolesGuard)
-  @Roles(Role.PROFESSOR)
+  @Get(':id/get_user_info/:role/:page/:search')
+  // @UseGuards(JwtAuthGuard, CourseRolesGuard)
+  // @Roles(Role.PROFESSOR)
   async getUserInfo(
     @Param('id') courseId: number,
     @Param('page') page: number,
     @Param('search') search: string,
-    @Param('role') role: Role,
+    @Param('role') role?: Role,
   ): Promise<UserPartial> {
     try {
-      return { id: courseId + page, name: search, email: role, photoURL: '' };
+      const users = await UserCourseModel.createQueryBuilder()
+        .innerJoin(
+          UserModel,
+          'UserModel',
+          '"UserModel".id = "UserCourseModel"."userId"',
+        )
+        .where('"UserCourseModel"."courseId" = :courseId', { courseId })
+        .andWhere('"UserCourseModel".role = :role', { role })
+        .getMany();
+      console.log('USERS');
+      console.log(users);
+      console.log('USERS');
+      console.log(users);
+      return {
+        id: page,
+        name: search,
+      };
     } catch (err) {
       console.error(err);
       throw new HttpException(
--- a/packages/server/src/course/course.controller.ts
+++ b/packages/server/src/course/course.controller.ts
@@ -596,12 +596,15 @@ export class CourseController {
     @Query('search') search?: string,
   ): Promise<UserPartial[]> {
     const pageSize = 50;
+    if (!search) {
+      search = '';
+    }
     const users = await this.courseService.getUserInfo(
       courseId,
       page,
       pageSize,
-      role,
       search,
+      role,
     );
     return users;
   }
--- a/packages/server/src/course/course.controller.ts
+++ b/packages/server/src/course/course.controller.ts
@@ -596,22 +596,14 @@ export class CourseController {
     @Query('search') search?: string,
   ): Promise<UserPartial[]> {
     const pageSize = 50;
-    try {
-      const users = await this.courseService.getUserInfo(
-        courseId,
-        page,
-        pageSize,
-        role,
-        search,
-      );
-      return users;
-    } catch (err) {
-      console.error(err);
-      throw new HttpException(
-        ERROR_MESSAGES.common.pageOutOfBounds,
-        HttpStatus.BAD_REQUEST,
-      );
-    }
+    const users = await this.courseService.getUserInfo(
+      courseId,
+      page,
+      pageSize,
+      role,
+      search,
+    );
+    return users;
   }
 
   @Post(':id/self_enroll')
--- a/packages/server/src/course/course.controller.ts
+++ b/packages/server/src/course/course.controller.ts
@@ -32,7 +32,7 @@ import async from 'async';
 import moment = require('moment');
 import { EventModel, EventType } from 'profile/event-model.entity';
 import { UserCourseModel } from 'profile/user-course.entity';
-import { Brackets, Connection, getRepository, MoreThanOrEqual } from 'typeorm';
+import { Connection, getRepository, MoreThanOrEqual } from 'typeorm';
 import { Roles } from '../decorators/roles.decorator';
 import { User, UserId } from '../decorators/user.decorator';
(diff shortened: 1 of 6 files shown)


packages/server/src/course/course.controller.ts
```diff
@@ -556,6 +557,34 @@ export class CourseController {
...
+  @UseGuards(JwtAuthGuard, CourseRolesGuard)
+  @Roles(Role.PROFESSOR)
+  async getUserInfo(
+    @Param('id') courseId: number,
+    @Param('page') page: number,
+    @Param('role') role?: Role,
+    @Query('search') search?: string,
+  ): Promise<UserPartial[]> {
+    const pageSize = 50;
+    try {
```
VKong6019, line 570: probably don't need this second try catch here if you have it at the service level

packages/server/src/course/course.service.ts
```diff
@@ -234,4 +236,63 @@ export class CourseService {
...
+          'UserModel.email',
+        ])
+        .orderBy('UserModel.firstName')
+        .skip((page - 1) * pageSize)
+        .take(pageSize)
+        .getMany();
+
+      return users;
+    } catch (err) {
+      console.error(err);
```
VKong6019, line 291: we can prob remove this console log here

packages/server/src/course/course.service.ts
```diff
@@ -234,4 +236,63 @@ export class CourseService {
...
+      throw new HttpException(
+        ERROR_MESSAGES.common.pageOutOfBounds,
```
VKong6019, line 293: hmm, will the error always be a `pageOutOfBounds`?

packages/server/src/course/course.service.ts
```diff
@@ -234,4 +236,63 @@ export class CourseService {
...
+      const users = query
+        .select([
+          'UserModel.id',
+          'UserModel.firstName',
+          'UserModel.lastName',
+          'UserModel.photoURL',
+          'UserModel.email',
+        ])
+        .orderBy('UserModel.firstName')
+        .skip((page - 1) * pageSize)
```
VKong6019, line 285: hmm, this isn't a suggestion for change. just food for thought: i suspect having `skip` won't be a problem in sql queries since I would hope there's not that many students/TAs in a course
https://stackoverflow.com/questions/34110504/optimize-query-with-offset-on-large-table/34291099#34291099

packages/server/src/course/course.service.ts
```diff
@@ -234,4 +236,63 @@ export class CourseService {
       );
     }
   }
+
+  async getUserInfo(
+    courseId: number,
+    page: number,
+    pageSize: number,
```
VKong6019, line 243: i think we can just have `pageSize` as a constant here right? idk if `getUserInfo` would be used anywhere else since this is specifically for the paginated endpoint

VKong6019 (commented):
sql query scary

also is this separate from `hybrid-queues` stuff

dfarooq610:
Yes this should exist separate of hybrid queues

dfarooq610:
Also reminder to write some tests 👀

packages/server/src/course/course.service.ts
sumitde22, line 293: To follow up on this, would this function ever throw an error? Just from a quick read, it seems like at worst it would return an empty list of users. Not 100% sure how take/skip behave, so feel free to correct me

packages/server/src/course/course.controller.ts
brianyyang: true, good catch!

packages/server/src/course/course.service.ts
brianyyang: no, @sumitde22 brings up a good point below too, I don't think it can ever be an out of bounds page anyways because the way the query is written it won't error out it will just return no students.

packages/server/src/course/course.service.ts
brianyyang: hmm i see so to summarize, `skip` is fine to use in our case because we don't have too much data in our tables, but if we did it would be very slow. therefore, if we wanted to optimize it we would paginate using `where` queries with a range of `id`'s instead?

packages/server/src/course/course.service.ts
brianyyang: that's what I had originally but I figured having page size associated with getting the user info might be good design for down the road but also idk. I guess this makes it so that you can change page size in the actual endpoint rather than in the service file?

packages/server/src/course/course.service.ts
brianyyang: I just tested it and I think you're right in the sense that the page can't go out of bounds. I'm not sure if other errors are possible though. It seems like if there are any malformed query parameters, the query would just return an empty list?

packages/server/src/course/course.controller.ts
brianyyang: o wait ima remove both bc i don't think it can error out

packages/server/src/course/course.service.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+        role?: Role,
+        search?: string,
+    */
+    it('returns nothing for a non-existing course id', async () => {
+      const courseId = 3;
+      const resp = await service.getUserInfo(courseId, page, pageSize);
+      expect(resp).toEqual([]);
+    });
+
+    it('returns Danish and Iris for fundies profs', async () => {
```
IrisLiu-00, line 170: i see that i have finally achieved immortalization in the KOH codebase via testing data 😤

packages/server/src/course/course.service.ts
IrisLiu-00: i think parametrizing the pageSize probably works in our favor for testing as well - like for those unit tests you can just set the page size to a really small number to check that pagination works

IrisLiu-00 (commented):
Nothing much to add here besides to complement the chungus work 🙏 💦

packages/server/src/course/course.service.ts
VKong6019: yessir

VKong6019 (approved):
Amazing

sumitde22 (approved):
Looks good to me

PR #860, opened 2022-03-13
Allow students to have multiple questions across multiple courses

### LINKED ISSUE #845: Allow students to have multiple questions across multiple courses
From a bug report:

> I was in one quese for two hours and wanted to join a queue for a different class for clarification and the other question is deleted.

The student was really frustrated, for good reason. We should allow students to have multiple questions across courses simultaneously.

tiingweii-shii:
# Description
Allows one question per course for each student.

Previously, only one question is allowed per user. We believe it is more ideal to allow a student to have one question per course. This PR also updated the warning for when a student wants to create a second question for a course.

Closes #845 

## Type of change

# How Has This Been Tested?

- [x] This has been manually tested in the front end. (In database: Create a second course in course_model. Add a student to the second course by adding an entry in user_course_model. On the app: Sign in as the student. Create a question in one of the courses that the student is in. Then create another question in the other course. This should not generate a warning. However, creating two questions in one course will give a warning.
- [x] One of the test in `question.controller.ts` has been adjusted to reflect that a second question in the same course (in different queues) gives a 400. A test has been added to show that a user can have one question each for two different courses.
- [x] This also has been tested with courses with more than one queues.

# Checklist:

### DIFF
--- a/packages/app/components/Queue/Student/StudentQueue.tsx
+++ b/packages/app/components/Queue/Student/StudentQueue.tsx
@@ -76,6 +76,10 @@ const HeaderText = styled.div`
   font-variant: small-caps;
 `;
 
+const PopConfirmTitle = styled.div`
+  max-width: 400px;
+`;
+
 const CenterRow = styled(Row)`
   align-items: center;
 `;
@@ -292,7 +296,13 @@ export default function StudentQueue({
             buttons={
               !studentQuestion && (
                 <Popconfirm
-                  title="In order to join this queue, you must delete your previous question. Do you want to continue?"
+                  title={
+                    <PopConfirmTitle>
+                      You already have a question in a queue for this course, so
+                      your previous question will be deleted in order to join
+                      this queue. Do you want to continue?
+                    </PopConfirmTitle>
+                  }
                   onConfirm={() => joinQueueOpenModal(true)}
                   okText="Yes"
                   cancelText="No"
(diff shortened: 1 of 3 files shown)


VKong6019 (approved):
yeehaw

packages/app/components/Queue/Student/StudentQueue.tsx
```diff
@@ -292,7 +296,13 @@ export default function StudentQueue({
             buttons={
               !studentQuestion && (
                 <Popconfirm
-                  title="In order to join this queue, you must delete your previous question. Do you want to continue?"
+                  title={
+                    <PopConfirmTitle>
+                      You already have a question in a queue for this course, so
```
dfarooq610, line 301: So we allow them to have multiple questions across the different courses. Do we still need this popup? 

Can't we just disable the button altogether? if they have a question in this course, then they can just edit their existing question. 

I think we can Disable the button and add a tooltip indicating that they can only have 1 question in the queue at a time

tiingweii-shii: Oh so we were thinking about hybrid queues when we implemented this.
Maybe there are students who want to leave an in-person queue for whatever reason and join an online queue instead (or vice versa)?

dfarooq610: Yea that makes sense, we can keep it in then

dfarooq610 (approved):

Issue #813, opened 2022-01-07
Use FullCalendar instead of BigCal on the FE


dfarooq610:
https://fullcalendar.io/docs/icalendar this component looks really handy. Since we no longer intend on scraping office hour blocks, we can just display whatever the course calendar does! Replace the big-cal components (on the schedule page and the today page) with full calendar, and try to use google calendar and outlook as event sources.

I don't think we can fully remove big-cal as a dependency just yet, unless you also want to modify check-in and check-out calendar in the admin page to use full calendar as well.

PR #861, opened 2022-03-23
Display OH schedule using FullCalendar

IrisLiu-00:
# Description

Reimplements the office hours schedule on the Today page and the Schedule page using the FullCalendar library. This library can display events directly from an iCal feed, so it is not dependent on our OfficeHoursModel.

Closes https://github.com/sandboxnu/office-hours/issues/813

Changes made:
* Add FullCalendar dependencies to do calendar display, swap out big-calendar from `SchedulePanel`. New calendar should have all features the old one had - you can switch between a month/week/day/agenda view, professors can update calendar thru a button
* Add nextjs plugin to prevent next from getting pissy at FullCal node modules importing css files 
* Create new `resources` controller with `calendar` endpoint to fetch ical data for a course. Fetching the ical data is done through the backend because if you try to directly request the outlook/google url from the frontend you get a CORS error
*ical data is saved to a file on the server to prevent refetching from google all the time. There is also a cron job that runs every day to refetch the file. 
* Created a google cal under the Sandbox gmail account for testing purposes - this is now the default calendar when you seed data, so you'll see calendar data from the front end

TODO:
- [ ] Should probably unit test the service methods somehow..?
- [ ] Probably should make a ticket to replace big-calendar in the TA checkin/checkout, so we can get rid of that dependency?
- [x] Test out the cron job
- [x] Prevent the cron job from running at the same time on both servers?

## Type of change

# How Has This Been Tested?

- [ ] Manually: navigate to Today and Schedule page and make sure the events show up. Edited some events (recurring + non recurring) and made sure the changes showed when the Update Calendar button was clicked. (Make sure to delete+seed data to get the new ical url in your db)
      - if you wanna test this with a heftier file, you can change the seeds to the the fundies cal: https://calendar.google.com/calendar/ical/5t8dq96sndpeu54813468o805g%40group.calendar.google.com/public/basic.ics

# Checklist:

# Screenshots:

image
image


packages/server/src/resources/resources.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+  Controller,
+  Get,
+  HttpService,
+  Inject,
+  Param,
+} from '@nestjs/common';
+import { Cache } from 'cache-manager';
+import { CourseModel } from '../course/course.entity';
+import { Connection } from 'typeorm';
+
```
dfarooq610, line 12: are you sure we wouldn't want a JWT token at least? People can only access this info if they are in an authorized session to use the app?

dfarooq610:
sometimes fundies will use the same calendar from semester to semester. meaning the ical file is huge. storing and cacheing requires a lot of memory allocation. I guess there is no other reasonable way to go about it though 😢

packages/server/src/resources/resources.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+@Controller('resources')
+export class ResourcesController {
+  constructor(
+    private connection: Connection,
+    private httpService: HttpService,
+    @Inject(CACHE_MANAGER) private cacheManager: Cache,
+  ) {}
+
+  @Get('/calendar/:course/refresh=:refresh')
```
dfarooq610, line 25: also can we then deprecate [this endpoint](https://github.com/sandboxnu/office-hours/blob/d14a6d9848c177d784ab82cadcc73a942ee3e08d/packages/server/src/course/course.controller.ts#L465) and update the cron job that re-scrapes the calendar endpoints? (or make a new cron job to parse and cache these at midnight

packages/server/test/util/factories.ts
```diff
@@ -46,7 +46,11 @@ export const OfficeHourFactory = new Factory(OfficeHourModel)
 
 export const CourseFactory = new Factory(CourseModel)
   .attr('name', 'CS 2500')
-  .attr('icalURL', 'http://hi.com')
+  // calendar is owned by sandboxneu@gmail.com
+  .attr(
+    'icalURL',
+    'https://calendar.google.com/calendar/ical/t6lu2pic7u9otrbpkuk26sl34g%40group.calendar.google.com/public/basic.ics',
```
dfarooq610, line 52: i should probably test if it works using an outlook url as well

packages/server/src/resources/resources.controller.ts
IrisLiu-00: if it's the fullcal lib making the request, would it have a JWT token? when i was writing this it seemed like putting the endpoint in any other controller caused guard issues

IrisLiu-00:
> sometimes fundies will use the same calendar from semester to semester. meaning the text from the ical file is huge. storing and caching it for all courses will require a lot of memory allocation. I guess there is no other reasonable way to go about it though.
> 
> I do not want to deal with parsing this ical string and excluding old events 😢

we could just not cache it :P

packages/server/test/util/factories.ts
IrisLiu-00: i tested with outlook and it seemed good, but feel free to test using your own calendars as well

packages/server/src/resources/resources.controller.ts
```diff
@@ -0,0 +1,41 @@
...
+  async getCourseCalendar(
+    @Param('course') courseId: number,
+    @Param('refresh') refresh: boolean,
+  ): Promise<string> {
+    const cacheKey = `/calendar/${courseId}`;
+    const cal = await this.cacheManager.get(cacheKey);
+    if (!cal || refresh) {
+      const course = await CourseModel.findOne(courseId);
+      const request = await this.httpService.get(course.icalURL).toPromise();
+      await this.cacheManager.set(cacheKey, request.data, { ttl: 86400 }); // cache results for 24 hrs
```
VKong6019, line 35: this is so cache money

dfarooq610 (approved):

PR #879, opened 2022-04-22
insert a clear queue button 

### LINKED ISSUE #878: Add Clear Queue functionality to queue page
Add a Clear Queue button to the queue page that cleans all the student questions from the queue. TAs can click it if they know that noone else is going to check in for a while.

* button should be marked as a danger button like the delete queue button is, although it might be good to place the button a bit more towards the top of the column since it's kind of a more commonly performed action? idk
* we should theoretically already have BE support for this

dfarooq610:
# Description
Button on queue page to allow TAs to clear the queue. Also fixing up some CSS (ty to Venkat for the assistance). 

https://user-images.githubusercontent.com/59672089/164585322-83325223-05f9-4db0-9326-e22e1562635c.mov

Closes #878 

## Type of change

## Testing
Manually tested -- backend endpoint is already tested. I could write a cypress test, but that may be a bit much.

### DIFF
--- a/packages/app/components/Queue/QueueListSharedComponents.tsx
+++ b/packages/app/components/Queue/QueueListSharedComponents.tsx
@@ -37,7 +37,6 @@ export const NotesText = styled.div`
 `;
 
 // New queue styled components start here
-
 const InfoColumnContainer = styled.div`
   flex-shrink: 0;
   padding-bottom: 30px;
@@ -99,13 +98,27 @@ const QueueRoomGroup = styled.div`
 const DisableQueueButton = styled(QueueInfoColumnButton)`
   color: white;
   background: #da3236;
-  bottom: 0;
-  position: absolute;
   &:hover,
   &:focus {
-    color: white;
-    background: #f76c6c;
+    color: #da3236;
+    background: #fff;
+    border-color: #da3236;
   }
+`;
+
+const ClearQueueButton = styled(QueueInfoColumnButton)`
+  background-color: #1890ff;
+  color: white;
+`;
+
+const QueueManagementBox = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  color: white;
+  width: 100%;
+  height: 100%;
+  bottom: 0;
 `;
 
 interface QueueInfoColumnProps {
@@ -123,9 +136,15 @@ export function QueueInfoColumn({
 
   const disableQueue = async () => {
     await API.queues.disable(queueId);
-    await mutateQueue(); // cope
+    await mutateQueue();
     message.success("Successfully disabled queue: " + queue.room);
     await Router.push("/");
+  };
+
+  const clearQueue = async () => {
+    await API.queues.clean(queueId);
+    await mutateQueue();
+    message.success("Successfully cleaned queue: " + queue.room);
   };
 
   const confirmDisable = () => {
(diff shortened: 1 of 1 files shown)


IrisLiu-00:
ooo niceeeeeee

I wonder if we should display a confirmation modal when you click on it, like we do for the delete queue button? It could say something like, don't clear the queue unless you're sure theres no more TAs after you

Also since this is kind of a more common action than fully deleting the queue, I wonder if it makes sense to put it more towards the top (like below the checkout button)? We could color it red like the delete queue button so it's clear that it's a danger action 🤔

dfarooq610:
i can put in a popconfirm

i think the button can go between the help next and check out buttons? i am not sure what to color it though since check out is already kinda red 🤔

dfarooq610:
although, because this is probably also rarely used, i actually kinda like keeping it down by the delete queue button

IrisLiu-00:
> although, because this is probably also rarely used, i actually kinda like keeping it down by the delete queue button

Really? I thought it would be somewhat regularly used, especially if a class doesn't have close set office hours throughout the week - like it might be used like once or twice a week?

You could do it fully red like the way the delete queue button is colored, maybe? 🤔 that'll mark it as a more "dangerous" action than just checking out is.

packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -123,11 +142,17 @@ export function QueueInfoColumn({
 
   const disableQueue = async () => {
     await API.queues.disable(queueId);
-    await mutateQueue(); // cope
```
VKong6019, line 126: LOL

packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -188,13 +213,16 @@ export function QueueInfoColumn({
...
       {isStaff && (
-        <DisableQueueButton
-          onClick={confirmDisable}
-          data-cy="queue-disable-button"
-          disabled={queue?.isDisabled}
-        >
-          {queue?.isDisabled ? `Queue deleted` : `Delete Queue`}
-        </DisableQueueButton>
+        <QueueManagementBox>
+          <ClearQueueButton onClick={clearQueue}>Clear Queue</ClearQueueButton>
```
VKong6019, line 217: +1 to iris on adding a popup confirmation modal for this

packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -99,15 +98,35 @@ const QueueRoomGroup = styled.div`
...
+
+const QueueManagementBox = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  color: white;
+  width: 100%;
+  height: 100%;
+  bottom: 0;
+`;
```
VKong6019, line 128: css necromancy? :0

packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -99,15 +98,35 @@ const QueueRoomGroup = styled.div`
...
-    color: white;
-    background: #f76c6c;
+    color: #da3236;
+    background: #fff;
+    border-color: #da3236;
   }
 `;
 
+const ClearQueueButton = styled(QueueInfoColumnButton)`
+  background-color: #ff8c00;
```
VKong6019, line 110: look at this vibrancy of this orange 😎 
image

dfarooq610:
orange + popconfirm

https://user-images.githubusercontent.com/59672089/165196365-a6729353-848b-49ef-af27-299639c71ff8.mov

VKong6019 (approved):
Yeer

VKong6019 (approved):
Yayeet

IrisLiu-00 (approved):
LGTM ✌️

PR #880, opened 2022-04-22
Tag for professor queues

### LINKED ISSUE #876: Display Prof queues using a tag
* On the Queue page, add some sort of tag or pill for Prof queues
* On the Today page, change QueueCards for prof queues to use that same pill (in place of displaying the word (Professor) next to the queue room)

sumitde22:
# Description

Adds a tag to indicate a professor queue to clean up UI a bit.

Closes #876

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Ran in dev and created a queue to check frontend update
Screen Shot 2022-04-24 at 3 52 02 PM

Screen Shot 2022-04-24 at 3 52 15 PM

# Checklist:

- [] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes


packages/app/components/Queue/QueueListSharedComponents.tsx
```diff
@@ -164,6 +164,12 @@ export function QueueInfoColumn({
           </QueuePropertyText>
         </QueuePropertyRow>
       )}
+
+      {queue?.isProfessorQueue && (
+        <QueuePropertyRow>
+          <Tag>Professor Queue</Tag>
+        </QueuePropertyRow>
+      )}
```
IrisLiu-00, line 172: Feels like there's a bit too much spacing between the queue title and this tag 🤔 it looks like the `QueueRoomGroup` above it has a sizeable bottom margin. You could get rid of the margin on that and group both the group and the tag together in a div, then set the bottom margin on that div?

Also I just realized there's that thing to display the queue start/end time right above this (line 159) which we don't need anymore - would you mind deleting it while you're here :,,)

packages/app/components/Today/OpenQueueCard.tsx
```diff
@@ -149,17 +149,18 @@ const OpenQueueCard = ({
         borderRadius: "6px 6px 0 0",
       }}
       className={"open-queue-card"}
-      title={
-        <span>
-          {queue.room} <i>{queue.isProfessorQueue ? `(Professor)` : ``}</i>{" "}
-        </span>
-      }
+      title={<span>{queue.room} </span>}
```
IrisLiu-00, line 152: I wonder if the tag would look better up in the header right here 🤔 putting it in the body of the card looks a bit weird to me, especially if you also close the queue to new questions - 
image

I think if you put the tag in the header, it might be nice to make the tag like light blue colored or something, so the contrast is a bit less jarring! Now that I think of it making it light blue in the Queue page would help it stand out a bit more too

IrisLiu-00 (commented):
Thanks for getting this!!

sumitde22:
@IrisLiu-00 Good suggestions all around, added some fixes and uploaded the new screenshots, let me know if they match what you were looking for

packages/app/components/Today/OpenQueueCard.tsx
```diff
@@ -167,17 +167,22 @@ const OpenQueueCard = ({
...
-                  color="error"
-                  style={{ margin: 0 }}
-                >
-                  Not Accepting Questions
-                </Tag>
-              </Tooltip>
-            )}
+            <QueueInfoTags>
+              {queue?.isProfessorQueue && (
+                <Tag color="blue">Professor Queue</Tag>
```
VKong6019, line 172: i think you'd want to include a `style={{ margin: 0 }}` here because there's a small indent here
image

VKong6019: might be worth abstracting that 🤔, if possible ofc

IrisLiu-00 (approved):
Thanks for getting those changes! ✨

packages/app/components/Today/OpenQueueCard.tsx
sumitde22: Good catch, fixed. By abstracting, do you mean like making that type of tag a styled component or something else

VKong6019: Oh ye smth like that, but i think that change would be a nit. So not a big deal imo

VKong6019 (approved):

PR #885, opened 2022-04-27
cannot hit rephrase alert if question status is drafting

dfarooq610:
# Description
minor case, TAs cant see student text while theyre drafting so it does not make sense for them to ask students to rephrase their question 

# How Has This Been Tested?
- [ ] it is very straightforward


IrisLiu-00 (approved):
yup yup

PR #894, opened 2022-05-17
Add logging to calendar cron job

IrisLiu-00:
# Description

For some reason, calendar files for old semesters aren't getting deleted on prod :/ I've confirmed the feature works locally though, so the only thing I can think of is to add a bunch more logs to try to see what's happening on prod.


ashdawngary (approved):

VKong6019 (approved):
You go hackermN

PR #899, opened 2022-05-25
Sandbox Application Ad

tiingweii-shii:
# Description

Added a Modal on the Today page to advertise Sandbox Application Opening, cause KOH is an advertising website.
(Special thanks to frontend god and goddess Danish and Nicky :D)

# How Has This Been Tested?
Tested manually and nothing broke 😳

# Checklist:

### DIFF
--- a/packages/app/components/Today/SandboxApplication.tsx
+++ b/packages/app/components/Today/SandboxApplication.tsx
@@ -1,0 +1,51 @@
+import { Modal, Button } from "antd";
+import { ReactElement } from "react";
+import { useLocalStorage } from "../../hooks/useLocalStorage";
+
+export default function ApplyToSandbox(): ReactElement {
+  const [firstTime, setFirstTime] = useLocalStorage("seenApplyModal", true);
+
+  return (
+    <Modal
+      visible={firstTime}
+      footer={null}
+      onCancel={() => setFirstTime(false)}
+      width={625}
+    >
+      <div>
+        <h1> Sandbox Applications Are Now Open! 🎉</h1>
+        <a
+          href={"https://www.sandboxnu.com/"}
+          target={"_blank"}
+          rel={"noreferrer"}
+        >
+          Sandbox
+        </a>{" "}
+        is Northeastern&#39;s student-led software consultancy that aims to
+        unleash the power of student-driven software. Not only do we engineer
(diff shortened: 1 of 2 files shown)


packages/app/components/Today/SandboxApplication.tsx
```diff
@@ -0,0 +1,51 @@
+import { Modal, Button } from "antd";
+import { ReactElement } from "react";
+import { useLocalStorage } from "../../hooks/useLocalStorage";
+
+export default function ApplyToSandbox(): ReactElement {
+  const [firstTime, setFirstTime] = useLocalStorage("seenApplyModal", true);
+
+  return (
+    <Modal
+      visible={firstTime}
```
dfarooq610, line 10: I wonder if we have some constants APPS_OPEN and APPS_CLOSE and we render if we are between those dates and it's the first time

(So people won't see it in summer 2 once apps are over)

packages/app/components/Today/SandboxApplication.tsx
```diff
@@ -0,0 +1,51 @@
...
+          target={"_blank"}
+          rel={"noreferrer"}
+        >
+          SearchNEU
+        </a>{" "}
+        and Khoury Office Hours.
+        <br />
+        <br />
+        <div style={{ justifyContent: "space-between", display: "flex" }}>
+          <div>Application closes June 5th 2022.</div>
```
dfarooq610, line 39: I think this should say "Applications close June 5th 2022"

dfarooq610 (commented):
Also, do you think it would be a good idea to add a button in the footer?

packages/app/components/Today/SandboxApplication.tsx
tiingweii-shii: ok i will fix this and you go sleep

IrisLiu-00 (approved):
LGTM im so ready to sell ourselves out as makers of soulless ad tech 💰

packages/app/components/Today/SandboxApplication.tsx
tiingweii-shii: Yeee I like this idea! But I probably don't have time to implement this until later (sometime in June) 😥

dfarooq610 (approved):

PR #905, opened 2022-06-04
Course roster page frontend

### LINKED ISSUE #866: Create Course Roster Page
Implement the designs for the course roster page within the course admin page using the endpoint from #843

sumitde22:
# Description

Add new course roster page to course admin panel. Lists professors, TAs, and students in course, and allows admin to search through these lists. Also offers pagination over student list

Also makes change to API so that it sends the total number of students/TAs/professors etc. along with the page of users requested

Closes #866

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

Tested to make sure API requests + display worked correctly on dev, added ~ 50 dummy students to test pagination/search features

Top of Page display 

Screen Shot 2022-06-04 at 3 46 21 PM

Bottom of Page display

Screen Shot 2022-06-04 at 3 46 47 PM

# Checklist:

- [] I have added tests that prove my fix is effective or that my feature works
- [] New and existing unit tests pass locally with my changes

### DIFF
--- a/packages/app/components/Settings/CourseRoster.tsx
+++ b/packages/app/components/Settings/CourseRoster.tsx
@@ -24,7 +24,7 @@ const CourseRosterComponent = styled.div`
 `;
 
 const TableBackground = styled.div`
-  backgroundcolor: white;
+  background-color: white;
 `;
 
 export default function CourseRoster({
--- a/packages/app/components/Settings/CourseRoster.tsx
+++ b/packages/app/components/Settings/CourseRoster.tsx
@@ -1,0 +1,122 @@
+import { SearchOutlined } from "@ant-design/icons";
+import { API } from "@koh/api-client";
+import { Role } from "@koh/common";
+import { Divider, Input, List, Pagination, Spin } from "antd";
+import Avatar from "antd/lib/avatar/avatar";
+import { useState } from "react";
+import { ReactElement } from "react";
+import styled from "styled-components";
+import useSWR from "swr";
+
+type CourseRosterProps = { courseId: number };
+
+type RenderTableProps = {
+  courseId: number;
+  role: Role;
+  listTitle: string;
+  displaySearchBar: boolean;
+  searchPlaceholder: string;
+};
+
+const CourseRosterComponent = styled.div`
+  margin-left: auto;
+  margin-right: auto;
+`;
+
+export default function CourseRoster({
+  courseId,
+}: CourseRosterProps): ReactElement {
+  return (
+    <CourseRosterComponent>
+      <h1>Course Roster</h1>
+      <RenderTable
+        courseId={courseId}
+        role={Role.PROFESSOR}
+        listTitle={"Professors"}
+        displaySearchBar={false}
+        searchPlaceholder=""
+      />
+      <br />
+      <RenderTable
+        courseId={courseId}
+        role={Role.TA}
+        listTitle={"Teaching Assistants"}
+        displaySearchBar={true}
+        searchPlaceholder="Search TAs"
+      />
+      <br />
+      <RenderTable
+        courseId={courseId}
+        role={Role.STUDENT}
+        listTitle={"Students"}
+        displaySearchBar={true}
+        searchPlaceholder="Search students"
+      />
+      <br />
+      <Divider />
+    </CourseRosterComponent>
+  );
+}
+
+function RenderTable({
+  courseId,
+  role,
+  listTitle,
+  displaySearchBar,
+  searchPlaceholder,
+}: RenderTableProps): ReactElement {
+  const [page, setPage] = useState(1);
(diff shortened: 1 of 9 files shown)


packages/app/components/Settings/CourseRoster.tsx
```diff
@@ -0,0 +1,122 @@
...
+function RenderTable({
+  courseId,
+  role,
+  listTitle,
+  displaySearchBar,
+  searchPlaceholder,
+}: RenderTableProps): ReactElement {
+  const [page, setPage] = useState(1);
+  const [search, setSearch] = useState("");
+  const handleSearch = (event) => {
```
IrisLiu-00, line 70: it looks like this is submitting an API request with every character you type, which based on my manual testing feels a bit rough - you have to type a character, wait for it to load, then type another character. I think we can probably make the UX a better by either debouncing the request (basically, set a timeout and only submit the request if x seconds pass without the user typing something), or just only submit the search request when the user hits the enter key. Hitting the enter key is probably the easier approach, but if you wanna do the debounce feel free to talk to me if you want help or anything :P

packages/app/components/Settings/CourseRoster.tsx
```diff
@@ -0,0 +1,122 @@
...
+    `${role}/${page}/${search}`,
+    async () => await API.course.getUserInfo(courseId, page, role, search)
+  );
+  if (!data) {
+    return <Spin tip="Loading..." size="large" />;
+  } else {
+    return (
+      <div style={{ backgroundColor: "white" }}>
+        <div style={{ backgroundColor: "#f0f0f0", height: "56px" }}>
+          <h3 style={{ position: "relative", left: "10px", top: "14px" }}>
```
IrisLiu-00, line 85: might be nice to put some of these as styled components :>

IrisLiu-00 (changes requested):
👏 👏 this looks awesome!

packages/app/components/Settings/CourseRoster.tsx
IrisLiu-00: oh, and I also noticed that it seems like you can search by people's names, but not by emails 🤔 I don't think it's too important, but if it's not difficult to make that change it might be nice to have as well

dfarooq610:
Something a bit strange is happening: i added about 20 students, TAs and profs to the db. When i go to the apply page, they are present in the requests, but arent rendered in the table, even after reloading 🤔 
image

packages/app/components/Settings/CourseRoster.tsx
```diff
@@ -0,0 +1,122 @@
...
+            value={search}
+            onChange={handleSearch}
+          />
+        )}
+        <List
+          dataSource={data.users}
+          renderItem={(item) => (
+            <List.Item key={item.id}>
+              <List.Item.Meta
+                avatar={<Avatar src={item.photoURL} />}
```
dfarooq610, line 102: it would be nice to reuse the AvatarWithInitals component (so if they dont have a photo we render their initials) 
image

packages/app/components/Settings/CourseRoster.tsx
```diff
@@ -0,0 +1,122 @@
...
+                title={item.name}
+              />
+              <div>{item.email}</div>
+            </List.Item>
+          )}
+          bordered
+        />
+        {data.total > 50 && (
+          <Pagination
```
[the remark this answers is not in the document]
sumitde22: Fixed

packages/app/components/Settings/CourseRoster.tsx
sumitde22: Good point, I took the enter approach as it seemed intuitive to me and made the change

Think I would have to make some changes to the backend to allow for email searching, will do that real quick

packages/app/components/Settings/CourseRoster.tsx
sumitde22: Got 1 styled component in, the second div doesn't keep its color when I try to change it to a styled component so I think I'll keep that as is

packages/app/components/Settings/CourseRoster.tsx
sumitde22: Just added that option

IrisLiu-00 (commented):
Changes are looking great! I just pushed a small commit fixing a typo - css properties are camelCase when they're part of the `style` prop, but they're hyphenated when you're writing direct css (like in a styled component). 

Depending on how much work implementing search by email would be, we could also leave it out - the emails are just made up of the person's first + last names, anyways

IrisLiu-00 (approved):

dfarooq610 (approved):

